PR #3, opened 2020-09-29
Fix authorization and json_body

songofroland:


tests/test_authz_Authorizer.py
```diff
@@ -1,11 +1,19 @@
...
+import random
+import string
+
 import pytest
 
 from lbz.authz import Authorizer
 from lbz.exceptions import PermissionDenied, ServerError
 from lbz.misc import NestedDict
 
+ALL = "*"
```
pdyba, line 12: You could and should import that form auth

lbz/response.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/local/bin/python3.8
```
pdyba, line 1: why did You split communication into 2 files ?

tests/test_authz_Authorizer.py
pdyba: that and below.

lbz/response.py
songofroland: To avoid circular imports

pdyba (approved):

PR #4, opened 2020-10-05
Add Cognito User module

songofroland:


lbz/authentication.py
```diff
@@ -0,0 +1,45 @@
...
+            raise RuntimeError
+
+    def __repr__(self):
+        if hasattr(self, "id"):
+            return f"User id={self.id}"
+        elif hasattr(self, "username"):
+            return f"User username={self.username}"
+
+
+def load_cognito_user(token: str, public_jwk: Union[dict, str], pool_id: str) -> User:
```
pdyba, line 31: this should be class method of User same below.

lbz/authentication.py
```diff
@@ -0,0 +1,45 @@
...
+        for k, v in kwargs.items():
+            self.__setattr__(k, v)
+
+    @staticmethod
+    def _validate_attributes(attributes: dict):
+        if len(attributes) > 1000:
+            logger.error(f"Too many attributes, total={len(attributes)}")
+            raise RuntimeError
+
+    def __repr__(self):
```
pdyba, line 24: __ methods should be before _

PR #17, opened 2020-12-06
Auth rework

grzegorzpro:


lbz/resource.py
```diff
@@ -84,4 +85,4 @@ def get_guest_authorization(self) -> str:
         It should be overwritten after inheritance with a method to obtain guest auth.
         """
         logger.error("Using default guest authorization which gives root access.")
-        return self._authorizer.sign_authz({"allow": {"*": "*"}, "deny": {}})
+        return self._authorizer.sign_authz({"allow": {"*": "*"}, "deny": {}}, sample_private_key)
```
songofroland, line 88: I would consider removing this guest auth, otherwise we'll need to overwrite it in every authorized api. I think it's more a security leak than a convenience tool - what do you think ? 

Especially now, as it will be signed by sample private key, we'll need to set sample public key (makes no sense for dev now).

songofroland (approved):

lbz/resource.py
grzegorzpro: Good catch, I thought this is some test code when doing basically find + replace for the `sign_authz` method. This should definitely not be here

lbz/authz.py
```diff
@@ -101,6 +101,15 @@ def set_policy(self, token: str):
         self.allowed_resource = None
         self.denied_resource = None
 
+    def reset_policy(self):
```
songofroland, line 104: There is method `set_initial_state` please remove it and insert this (they have same purpose)

songofroland (approved):

lbz/authz.py
grzegorzpro: I'm not sure about this, I think `set_initial_state` is supposed to prepare the authorizer for permissions check (so it also sets `self.action`, while this `reset_policy` is supposed to remove all information about the policy because the request didn't have any authorization data.
Anyway I want to redesign this class, I think the way the authorization is executed can be simplified a lot and afterwards these methods won't be necessary anymore.

pdyba (approved):

PR #19, opened 2020-12-17
Multiple enhancements

songofroland:
* Headers as case-insensitive dict,
* "urn" attribute for Resource class,
* auth flags deprecation.


lbz/request.py
```diff
@@ -65,10 +61,10 @@ def raw_body(self) -> Union[bytes, str]:
         return self._raw_body
 
     @property
-    def json_body(self) -> dict:
-        content_type = self.headers.get("Content-Type", self.headers.get("content-type"))
+    def json_body(self) -> Union[dict, None]:
```
grzegorzpro, line 64: You can also do `Optional[dict]` :)

tests/test_misc.py
```diff
@@ -33,23 +26,21 @@ class A(metaclass=Singleton):
...
+    assert multi_dict.getlist("a") == ["a", "b", "c"]
+
+    del multi_dict["a"]
+    assert multi_dict.get("a") is None
+    assert multi_dict.__repr__() == str(multi_dict)
+    assert str(multi_dict) == "MultiDict({})"
+    multi_dict["b"] = "abc"
     acc = 0
-    for a in x:
+    for _ in multi_dict:
```
grzegorzpro, line 40: I'd say the dict should have a `len` method, and if we want to test the iteration, we should also check the values

grzegorzpro (approved):

PR #20, opened 2020-12-18
Exceptions reworked:

songofroland:
* added request id,
* simplified LambdaFWException class.

### DIFF
--- a/lbz/exceptions.py
+++ b/lbz/exceptions.py
@@ -2,6 +2,8 @@
 # coding=utf-8
 from http import HTTPStatus
 
+from lbz.response import Response
+
 
 class LambdaFWException(Exception):
     message = HTTPStatus.INTERNAL_SERVER_ERROR.description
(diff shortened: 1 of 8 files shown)


lbz/exceptions.py
```diff
@@ -2,32 +2,15 @@
...
     status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
-    headers = {}
-    resp = None
 
-    def __init__(self, message=""):
+    def __init__(self, message: str = ""):
         if message:
             self.message = message
 
-    def get_resp(self):
```
grzegorzpro, line 18: I'd actually leave this method, it is heavily related to this class and it might come in handy in some weird situation. Maybe rename it to `get_lbz_response` and remove `self.resp` (so just `return Response(...)`)

tests/test_resource.py
```diff
@@ -57,12 +58,7 @@ def setup_method(self):
     def test___call__wrong_uri(self):
         resp = self.res()
         assert isinstance(resp, Response)
-        assert resp.to_dict() == {
-            "body": '{"message":"Server is not able to produce a response"}',
-            "headers": {},
-            "statusCode": 421,
-            "isBase64Encoded": False,
-        }
+        assert resp.status_code == HTTPStatus.MISDIRECTED_REQUEST
```
grzegorzpro, line 61: Shouldn't this be just 404? This 421 code doesn't even have a page at MDN, I'd be really scratching my head if I saw it somewhere

grzegorzpro (approved):

pdyba (approved):

PR #21, opened 2020-12-18
Rework the authz module - remove singleton, use only one decorator

grzegorzpro:


lbz/authz.py
```diff
@@ -25,60 +21,25 @@
...
-            )
-        if self.iss is None:
-            warnings.warn(
-                "Lack ALLOWED_ISS is a security risk You should add it.",
-                SecurityRiskWarning,
-            )
-        elif self.iss and self.iss != ALLOWED_ISS:
-            raise PermissionDenied(f"{self.iss} is not allowed token issuer")
-        self.set_initial_state(function_name)
         if self.deny:
```
songofroland, line 43: This attribute is tricky for me to spot, because declared outside of ` __init__`.

lbz/authz.py
```diff
@@ -87,44 +48,46 @@ def validate(self, function_name):
...
-    def set_policy(self, token: str):
-        policy = decode_jwt(token)
-        self.allow = policy["allow"]
-        self.deny = policy["deny"]
-        self.expiration = policy.get(EXPIRATION_KEY)
-        self.iss = policy.get("iss")
-        self.outcome = DENY
-        self.allowed_resource = None
-        self.denied_resource = None
+    def _set_policy(self, auth_jwt: str):
```
songofroland, line 51: Could you please move it right after the `__init__` ? I think that will make readeing the code faster.

lbz/authz.py
```diff
@@ -148,12 +111,13 @@ def _check_allow(self):
             if d_domain := self.allow.get(self.resource):
                 if self._allow_if_allow_all(d_domain):
                     return
-                elif resource_to_check := d_domain.get(self.action):
+                elif resource_to_check := d_domain.get(self.permission):
                     self.outcome = ALLOW
                     self.allowed_resource = resource_to_check.get("allow")
                     self.denied_resource = resource_to_check.get("deny")
```
songofroland, line 118: Could you please rename `check_allow` to `check_allow_and_set_resources` ? IMHO this method needs to be redesigned, but for now I would at least make it clear that this function does not only one thing (which might be helpfull for debugging).

lbz/resource.py
```diff
@@ -44,10 +42,6 @@ def __init__(self, event: dict):
             is_base64_encoded=event.get("isBase64Encoded", False),
             query_params=event["multiValueQueryStringParameters"],
         )
-        if authorization := headers.get("Authorization"):
```
songofroland, line 47: Fantastic job with this refactor, big thanks.

songofroland (approved):

lbz/authz.py
grzegorzpro: I'll move it after `__repr__`, I think it's good to keep the `__x__` methods in the beginning of a class

lbz/authz.py
```diff
@@ -25,106 +21,74 @@
...
-        self._permissions = NestedDict()
-
-    def __getitem__(self, y):
-        return self._permissions[y]
-
-    def __str__(self):
-        return json.dumps({self.resource: self._permissions})
+class Authorizer:
+    def __init__(self, auth_jwt: str, resource_name: str, permission_name: str):
+        self.outcome = DENY
```
pdyba, line 26: its better to have them as class variables as they are not being changed during init.

lbz/authz.py
```diff
@@ -25,106 +21,74 @@
...
     def __repr__(self):
-        return self.__str__()
-
-    def __contains__(self, *args, **kwargs):
-        return self._permissions.__contains__(*args, **kwargs)
-
-    def __len__(self):
-        return len(self._permissions)
-
-    def __iter__(self):
```
pdyba, line 57: why remove all above ?

lbz/resource.py
pdyba: +1

pdyba (approved):

lbz/authz.py
grzegorzpro: I'd say logically they should be declared in init - they are attributes of an instance, not of the whole class. Declaring instance attributes as class attributes can lead to problems, I don't remember specific cases now, but I know I had to deal it at some point in the past.

grzegorzpro: Ok, here is an example:
```
class Abc:
    aaa = []

    def inc(self):
        self.aaa.append(1)

q = Abc()
w = Abc()
print(w.aaa)
q.inc()
print(w.aaa)
```
the second print will print `[1]` which is not what we expect since we never run `inc()` on `w`

lbz/authz.py
grzegorzpro: - class attributes were converted to instance attributes since the class is not a singleton anymore
- I removed the `_permissions` dict since it seemed unnecessary, most of these deleted functions were related to that dict so they were removed too

lbz/authz.py
pdyba: ACK

PR #23, opened 2020-12-22
reinstates guest

pdyba:


lbz/authz.py
```diff
@@ -22,24 +22,27 @@
 
 
 class Authorizer:
-    def __init__(self, auth_jwt: str, resource_name: str, permission_name: str):
+    def __init__(self, auth_jwt: str, resource_name: str, permission_name: str, auth_scope: dict = None):
```
songofroland, line 25: I don't see any line where `auth_scope` is added when Authorizer instantiated

grzegorzpro, line 25: I'd rename `auth_scope` -> `policy`, or maybe `policy_override` to be clear that this overrides the policy from the jwt

lbz/authz.py
```diff
@@ -138,13 +141,17 @@ def decorator(func: Callable):
         @wraps(func)
         def wrapped(self, *args, **kwargs):
             authorization_header = self.request.headers.get("Authorization")
-            if not authorization_header:
+            authorization_scope = None
+            if hasattr(self, "get_guest_authorization"):
```
grzegorzpro, line 145: So if the resource has a `get_guest_authorization` method it will never use the policy from the auth header? I don't think that's right?

Also I think we should add a test case to the `test_authz_decorator` file

lbz/authz.py
```diff
@@ -22,24 +22,27 @@
...
     def __repr__(self):
         return (
             f"Authorizer(auth_jwt=<jwt>, resource_name='{self.resource}', "
             f"permission_name='{self.permission}')"
         )
 
-    def _set_policy(self, auth_jwt: str):
-        policy = decode_jwt(auth_jwt)
+    def _set_policy(self, auth_jwt: str, auth_scope: dict = None):
+        if auth_scope:
```
grzegorzpro, line 42: I'd do `policy = auth_scope if auth_scope is not None else decode_jwt(auth_jwt)` (not sure what exactly should happen if `auth_scope == {}`)

tests/test_authz_decorator.py
```diff
@@ -38,6 +38,25 @@ def handler(self, restrictions):
         resp = X({**event, "headers": {}})()
         assert resp.status_code == HTTPStatus.UNAUTHORIZED
 
+    @patch.dict(environ, env_mock)
+    def test_no_auth_header_guest_in_place(self, *args):
+        class X(Resource):
+            @add_route("/")
+            @authorization()
+            def handler(self, restrictions):
+                return Response("x")
```
grzegorzpro, line 47: Let's also check the restrictions here to make sure they match the guest policy

grzegorzpro (approved):

PR #26, opened 2021-01-16
Add refs feature to the Authorizer module

grzegorzpro:


tests/test_authz_authorizer.py
```diff
@@ -106,6 +106,30 @@ def test_validate_fail_one_scope(self):
         assert authorizer.outcome == LIMITED_ALLOW
         assert authorizer.restrictions == {"allow": None, "deny": "self"}
 
+    def test_refs(self):
+        restrictions = {"allow": "*", "deny": "example"}
+        authorizer = self._make_authorizer(
+            {
+                **self.token_payload,
+                "refs": {"api-access": restrictions},
+                "allow": {"res": {"permission_name": {"ref": "api-access"}}},
```
songofroland, line 115: What's `res`?

tests/test_authz_authorizer.py
```diff
@@ -106,6 +106,30 @@ def test_validate_fail_one_scope(self):
         assert authorizer.outcome == LIMITED_ALLOW
         assert authorizer.restrictions == {"allow": None, "deny": "self"}
 
+    def test_refs(self):
+        restrictions = {"allow": "*", "deny": "example"}
+        authorizer = self._make_authorizer(
+            {
+                **self.token_payload,
```
songofroland, line 113: Maybe remove the `**self.token_payload` and declare the refs explicitly here? (to not jump over the code) This is a little bit magic and may be hard to understand for someone new, but it's just my opinion.

lbz/authz.py
```diff
@@ -115,9 +118,17 @@ def _check_allow_and_set_resources(self):
                 if self._allow_if_allow_all(d_domain):
                     return
                 elif resource_to_check := d_domain.get(self.permission):
+                    effective_permissions = resource_to_check
+                    if ref_name := resource_to_check.get("ref"):
+                        if ref_name not in self.refs:
```
songofroland, line 123: 3 nested `ifs` in `try/catch` - could you please try to flatten this? If you don't have time or feel like it's not crucial then leave it but imho this will make debugging kinda hard.

songofroland (approved):

tests/test_authz_authorizer.py
pdyba: resource

songofroland: So maybe `my_resource` or `test_resource`

tests/test_authz_authorizer.py
grzegorzpro: `**self.token_payload` is here to add the common fields that are basically always the same in the tests (`exp`, `iss`, `iat`), I'm not sure if declaring them explicitly here makes anything clearer?

lbz/authz.py
grzegorzpro: Moved to a separate function

tests/test_authz_authorizer.py
songofroland: Allrgiht I prolly missed that refs are defined here in 120.

songofroland (approved):

PR #27, opened 2021-02-28
Default str for jsond dumos

pdyba:


grzegorzpro (approved):

songofroland:
I'd prefer to keep raising 500 instead of serializing objects this way. I don't see a case when I'd like to dump an object implicitly.

PR #31, opened 2021-05-27
Permission utils & misc refinements

songofroland:
- Remove deprecated parts of the repository 
- Add `has_permission` helper function
- Add common imports to the top level `__init__.py`


lbz/misc.py
```diff
@@ -89,18 +88,3 @@ def format_error(error):
 
 
 logger = get_logger(__name__)
-
```
pdyba, line 92: why remove it ?

pdyba (approved):

PR #33, opened 2021-06-13
Improvement made to the messages being logged by lib

redlickigrzegorz:
**Most important changes:**
1. Replace the custom `format_error()` method with a standard `exception()` one:
    - this change does not break anything - still, the `traceback` will be visible in logs but fortunately, not as a message that would be trimmed e.g. by Sentry
2. Move responsibility for formatting messages to the login system
3. Correct the messages sent to the logger when handling exceptions
    - some of the error logs have been replaced by warnings as the related situations don't require that level of logging
    - a few logs have been put into the exception as message (handled level higher) - thanks to that error logs are not repeated
    - one `Unauthorized` (treated as client problem) error has been replaced by `RuntimeError` (treated as server problem)


lbz/authentication.py
```diff
@@ -37,8 +34,6 @@ def get_user_details_from_auth_token(self) -> dict:
         return parsed_user
 
     def _validate_attributes(self, attributes: dict) -> None:
-        if len(attributes) > self._max_attributes:
-            logger.error(
```
redlickigrzegorz, line 41: FYI: This error log would be repeated a level higher when handling a `RuntimeError`.

lbz/exceptions.py
```diff
@@ -16,6 +16,9 @@ def __init__(self, message: str = ""):
         if message:
             self.message = message
 
+    def __str__(self) -> str:
+        return f"{self.status_code} {self.message}"
```
redlickigrzegorz, line 20: This is a string that will be visible in logs, if you have a better idea, feel free to share it with me 🙂

lbz/jwt_utils.py
```diff
@@ -26,16 +26,16 @@ def get_matching_jwk(auth_jwt_token: str) -> dict:
             if key["kid"] == kid_from_jwt_header:
                 return key
 
-        logger.error(f"Required key not found in configuration.")
```
redlickigrzegorz, line 29: FYI: I don't think it is worth having error logs for this situation (client error).

lbz/jwt_utils.py
```diff
@@ -26,16 +26,16 @@ def get_matching_jwk(auth_jwt_token: str) -> dict:
...
     except (JWTError, KeyError):
         raise Unauthorized
 
 
 def decode_jwt(auth_jwt_token: str) -> dict:
     if not PUBLIC_KEYS:
-        logger.error("Invalid configuration - no keys in the ALLOWED_PUBLIC_KEYS env variable")
-        raise Unauthorized
+        msg = "Invalid configuration - no keys in the ALLOWED_PUBLIC_KEYS env variable"
+        raise RuntimeError(msg)
```
redlickigrzegorz, line 40: FYI: I have replaced it with `RuntimeError` because it is the server-side problem. We can create even a custom exception for things like that if you want 😉

lbz/jwt_utils.py
```diff
@@ -47,8 +47,8 @@ def decode_jwt(auth_jwt_token: str) -> dict:
             raise Unauthorized(f"Your token has expired. Please refresh it.")
         except JWTError:
             raise Unauthorized
-        except Exception:
-            logger.error(f"Error during decoding, token={auth_jwt_token}")
-            raise ServerError
+        except Exception as ex:
+            msg = f"Error during decoding, token={auth_jwt_token}"
+            raise RuntimeError(msg) from ex
```
redlickigrzegorz, line 54: FYI: It will be handled a level higher as `ServerError` but the log is not repeated anymore.

lbz/exceptions.py
songofroland: How about braces for status code?

`return f"[{self.status_code}] {self.message}"`

lbz/jwt_utils.py
songofroland: I'd also adjust the message a bit:

`"The key with id={kid_from_jwt_header} was not found in the environment variable."`

lbz/jwt_utils.py
songofroland: I'd adjust the message in passing :wink: 

`An error occurred during decoding the token. \nToken body: \n{auth_jwt_token}` or equivalent

lbz/request.py
```diff
@@ -72,12 +73,10 @@ def json_body(self) -> Optional[dict]:
                 try:
                     self._json_body = json.loads(self.raw_body)
                 except ValueError:
-                    logging.error(f"Invalid json payload: {self.raw_body}")
-                    raise BadRequestError
+                    raise BadRequestError(f"Invalid json payload: {self.raw_body}")
```
songofroland, line 76: How about `The provided payload is invalid.\n Payload body: \n{self.raw_body}`

songofroland (approved):

lbz/exceptions.py
pdyba: Please adjust based on Rolands suggestion

pdyba (approved):

pdyba:
changes based on Rolands input and it can be merged.

PR #36, opened 2021-07-23
Exceptions cleanup and adds missing

pdyba:


lbz/authz.py
```diff
@@ -43,6 +43,9 @@ def __init__(
         self.allow: dict = {}
         self.deny: dict = {}
         self._set_policy(auth_jwt, policy_override)
+        self.error = PermissionDenied(
+            log_warn=f"You don't have permission to {self.permission} on {self.resource}"
```
redlickigrzegorz, line 47: I don't think it should be marked as a warning 🤔  The situations that the user is not privileged enough to get access somewhere are very common and a large number of logs of this level can make it difficult to find the more relevant ones.

I can agree with logging it as a debug, just to have information about that in logs 😉

lbz/exceptions.py
```diff
@@ -18,10 +18,14 @@ class LambdaFWException(Exception):
     message = HTTPStatus.INTERNAL_SERVER_ERROR.description
     status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
 
-    def __init__(self, message: str = ""):
+    def __init__(self, message: str = None, log_msg: str = None, log_warn: str = None):
```
redlickigrzegorz, line 21: I know it was moved from the `ServerError` but it is not the best way how messages should be logged.

A couple of reasons of why:
1. Each module should contain its own logger instance, just to have an easy way to get knowledge of where the log comes from. By doing that this way, all error/warning logs will be marked as from the exceptions module.
2. In most cases, the error logs should be registered on Exceptions handling, not on raising them. Of course, it depends on the situation in which the error appears but it is very possible that this feature won't be used that often.
3. The messages to the logger shouldn't be passed as a formatted string (rather as a message and its arguments) - by doing that this way, we lose that possibility.
4. The line with logger is that short that it shouldn't be a problem to write it in a specific place. From my perspective, this feature is not that beneficial (does not save duplicating a lot of code).

redlickigrzegorz: That is a good moment to extend the `__init__` with the arguments for the base class 😉 
```python
def __init__(self, message: str = None, *args, **kwargs):  # type: ignore
    super().__init__(*args, **kwargs)  # type: ignore
```

lbz/exceptions.py
```diff
@@ -37,83 +41,288 @@ def get_response(self, request_id: str) -> Response:
...
+
+
+class ServerError(LambdaFWException):
+    """500 - Server got itself in trouble"""
+
+    message = HTTPStatus.INTERNAL_SERVER_ERROR.description
+    status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
+
+
+class NotImplementedHTTP(LambdaFWException):
```
redlickigrzegorz, line 256: That is the only exception with an _HTTP_ suffix 🤔  Do we need to add that here?

tests/test_exceptions.py
```diff
@@ -21,55 +53,62 @@ def test_lambda_fw_exception():
...
+        URITooLong,
+        Unauthorized,
+        UnavailableForLegalReasons,
+        UnprocessableEntity,
+        UnsupportedMediaType,
+        UpgradeRequired,
+        VariantAlsoNegotiates,
+    ],
+)
+def test_all_exceptions(an_excetption):
```
redlickigrzegorz, line 105: The name of the test should point to a singular exception as it is parameterized and run multiple times e.g. `test_custom_exception()` 😉

tests/test_exceptions.py
```diff
@@ -21,55 +53,62 @@ def test_lambda_fw_exception():
...
     assert exp.message == "Unsupported method: GET"
     assert exp.status_code == 405
 
 
-def test_unauthorized():
-    exp = Unauthorized()
-    assert exp.message == "No permission -- see authorization schemes"
-    assert exp.status_code == 401
+@pytest.mark.parametrize(
+    "an_excetption",
```
redlickigrzegorz, line 63: Typo: _"exception"_ but also, I think it is not that popular to add articles into the variable names. You can name it `ex` or `error` or `custom_exception` or even `exception` shouldn't cause problems (there are a lot of options) 😉

tests/test_exceptions.py
```diff
@@ -21,55 +53,62 @@ def test_lambda_fw_exception():
...
+        UnsupportedMediaType,
+        UpgradeRequired,
+        VariantAlsoNegotiates,
+    ],
+)
+def test_all_exceptions(an_excetption):
+    exp = an_excetption()
+    code, msg = exp.__doc__.split(" - ")
+    assert exp.message == msg.strip()
+    assert exp.status_code == int(code)
```
redlickigrzegorz, line 109: You can also check if the exception is a subclass of `LambdaFWException`, just to force things like that 😉

lbz/exceptions.py
pdyba: Yest as we would overwrite NotImplemented builtin

redlickigrzegorz: It makes a lot of sense! So maybe we can name it e.g. [NotImplementedFunctionality](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501)? Just an idea but this HTTP sounds like it should be put everywhere or nowhere 😉

lbz/authz.py
```diff
@@ -43,6 +43,9 @@ def __init__(
         self.allow: dict = {}
         self.deny: dict = {}
         self._set_policy(auth_jwt, policy_override)
+        self.error = PermissionDenied(
```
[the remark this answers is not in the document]
pdyba: still seems better the duplicating the same code 4 times.

lbz/authz.py
pdyba: adjusted. this logging make debugging any auth issues much simpler

lbz/exceptions.py
pdyba: args are fine Exception._iinit__ does not take **kwargs ;) but otherway around adjusted.

lbz/authz.py
redlickigrzegorz: I think the same but I was only worried about the logging level 😉

lbz/exceptions.py
redlickigrzegorz: That's right - only PyCharm's stubs are saying something else. Anyway, I wanted to stay compatible with builtins 😉

lbz/authz.py
```diff
@@ -76,6 +76,10 @@ def _set_policy(self, auth_jwt: str, policy_override: dict = None):
         elif issuer != ALLOWED_ISS:
             raise PermissionDenied(f"{issuer} is not an allowed token issuer")
 
+    def _raise_error(self):
```
redlickigrzegorz, line 79: You can be more direct and name it `_raise_permission_denied()` 😉  It will simplify reading the code in the future as you won't have to check implementation to be sure what will be raised.

lbz/exceptions.py
```diff
@@ -13,13 +13,14 @@
 class LambdaFWException(Exception):
     """
     Standarised for AWS Lambda exception class.
+    Log Levels: CRITICAL, ERROR, WARNING, INFO, DEBUG
```
redlickigrzegorz, line 16: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings
> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

You skipped a blank line 😉

redlickigrzegorz: What does it mean? The exception itself does not contain logging functionality 🤔 

Also, the `logger` that is declared above can be removed.

lbz/exceptions.py
```diff
@@ -13,13 +13,14 @@
...
     """
 
     message = HTTPStatus.INTERNAL_SERVER_ERROR.description
     status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
 
-    def __init__(self, message: str = ""):
-        super().__init__()
+    def __init__(self, *args, message: str = None):
```
redlickigrzegorz, line 22: That change needs to be checked carefully during updating Lambdalizator in our services 🤔 

I can imagine, we are passing the message as a positional argument, and problems related to that won't be caught automatically (at least not everywhere).

lbz/exceptions.py
```diff
@@ -33,87 +34,291 @@ def get_response(self, request_id: str) -> Response:
         return Response(
             {"message": self.message, "request_id": request_id},
             status_code=self.status_code,
-            headers={"Content-Type": "application/json"},
```
redlickigrzegorz, line 36: Why did you remove that? Or what am I skipping around this change? 🤔

tests/test_exceptions.py
```diff
@@ -1,75 +1,119 @@
...
+        UpgradeRequired,
+        VariantAlsoNegotiates,
+    ],
+)
+def test_custom_exception(custom_exception):
+    exp = custom_exception()
+    try:
+        code, msg = exp.__doc__.split(" - ")
+    except ValueError:
+        code, msg = exp.__doc__.strip(" -"), ""
```
redlickigrzegorz, line 110: I think, more cleaner is to update docstrings (they look more consistent thanks to that too) e.g.:
```python
"""426 - """
#       ^ empty sign
```

pdyba: that was initial way - black removes that

lbz/exceptions.py
pdyba: its added by default - unnecesery

lbz/exceptions.py
pdyba: fixed - got deep with base exception class

tests/test_exceptions.py
redlickigrzegorz: That's sad 😢  but we agreed to use `black` we need to accept that too.

lbz/exceptions.py
```diff
@@ -18,8 +15,8 @@ class LambdaFWException(Exception):
     message = HTTPStatus.INTERNAL_SERVER_ERROR.description
     status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
 
-    def __init__(self, message: str = ""):
-        super().__init__()
+    def __init__(self, message: str = "", errors: list = None):
```
redlickigrzegorz, line 18: I wonder only where did you find that Exception accepts the message and errors 🤔  Can you share documentation or something? I see in [typeshed/builtins.pyi](https://github.com/python/typeshed/blob/7c37d6117d0141c3cabd2880e160410308f4c060/stdlib/builtins.pyi#L1378) that it just accepts `*args`, nothing else.

tests/test_exceptions.py
redlickigrzegorz: You can split it and thanks to that you will get an empty string automatically: `exp.__doc__.split(" -")` 😉

redlickigrzegorz (approved):

PR #37, opened 2021-07-23
adds mypy and flake8

pdyba:


lbz/authentication.py
```diff
@@ -15,7 +15,7 @@ def remove_prefix(text: str):
     """
     Removes prefix of a text based on : sign.
     """
-    return text[text.index(":") + 1 :] if ":" in text else text
+    return text[text.index(":") + 1 :] if ":" in text else text  # noqa:E203
```
songofroland, line 18: I'm wondering if sth like `re.sub(".*\:", "", text)` would be enough, instead of the function

lbz/authz.py
```diff
@@ -166,31 +166,28 @@ def sign_authz(authz_data: dict, private_key_jwk: dict) -> str:
...
     """
-    authorization_header = resource.request.headers.get("Authorization")
-    authorization_scope = None
-    if not authorization_header:
-        if hasattr(resource, "get_guest_authorization"):
-            authorization_scope = resource.get_guest_authorization()
-        else:
-            raise Unauthorized("Authorization header missing or empty")
+    authorization_header = resource.request.headers.get("Authorization", "")
+    authorization_scope = resource.get_guest_authorization()
```
songofroland, line 176: I think `guest_authorization_policy` instead of `authorization_scope` would be less confusing

What if the resource does not have this method implemented?

lbz/authz.py
```diff
@@ -166,31 +166,28 @@ def sign_authz(authz_data: dict, private_key_jwk: dict) -> str:
...
+        resource_name=resource.get_name(),
         permission_name=permission_name,
         policy_override=authorization_scope,
     )
     authorizer.check_access()
     return authorizer.restrictions
 
 
-def has_permission(resource: Resource, permission_name: str) -> bool:
+def has_permission(resource: Union[Type[Resource], Resource], permission_name: str) -> bool:
```
songofroland, line 190: Is there really a case such that the class object is passed instead of an instance?

lbz/dev/event.py
```diff
@@ -0,0 +1,52 @@
+# flake8: noqa
+EVENT = """
```
songofroland, line 2: maybe `SAMPLE_EVENT`? @redlickigrzegorz

lbz/dev/server.py
```diff
@@ -5,22 +5,27 @@
 import json
 import logging
 import urllib.parse
+from abc import ABCMeta, abstractmethod
```
songofroland, line 8: wrong import order

lbz/dev/server.py
```diff
@@ -5,22 +5,27 @@
...
     Mimics AWS Lambda behavior.
     """
 
-    cls = None
-    done = False
+    done: bool = False
+
+    @property
+    @abstractmethod
+    def cls(self) -> Type[Resource]:
```
songofroland, line 27: WTF

lbz/misc.py
```diff
@@ -114,7 +115,7 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
-def copy_without_keys(data: dict, *keys) -> dict:
+def copy_without_keys(data: Union[dict, MultiDict], *keys) -> dict:
```
songofroland, line 118: I think it could be done better, by accepting all dict subclases, not sure tho

lbz/request.py
```diff
@@ -75,8 +75,9 @@ def json_body(self) -> Optional[dict]:
                 try:
                     self._json_body = json.loads(self.raw_body)
                 except ValueError as error:
-                    msg = f"The provided payload is invalid.\nPayload body:\n{self.raw_body}"
-                    raise BadRequestError(msg) from error
+                    raise BadRequestError(
+                        "Payload is invalid.\nPayload body:\n {!r}".format(self.raw_body)
```
songofroland, line 79: missing definite article

lbz/resource.py
```diff
@@ -99,6 +103,12 @@ def post_request_hook(self):
         Place to configure post request hooks.
         """
 
+    @staticmethod
+    def get_guest_authorization() -> dict:
+        """
+        Place to configure default authorization.
```
songofroland, line 109: it's a bit vague for me

lbz/resource.py
```diff
@@ -99,6 +103,12 @@ def post_request_hook(self):
         Place to configure post request hooks.
         """
 
+    @staticmethod
```
songofroland, line 106: it could be a good idea to make it `abstractmethod`

tests/__init__.py
```diff
@@ -3,7 +3,11 @@
 import os
 from uuid import uuid4
 
-from tests.fixtures.rsa_pair import sample_private_key, sample_public_key
+from tests.fixtures.rsa_pair import (  # noqa:F401
```
songofroland, line 6: what this `noqa` is for here?

tests/fixtures/cognito_auth.py
```diff
@@ -1,5 +1,5 @@
 env_mock = {
     "COGNITO_AUTHENTICATION": "1",
-    "ALLOWED_PUBLIC_KEYS": '{"keys": [{"kid": "12345", "n": "xyz"}, {"kid": "54321", "n": "xyz"}]}',
+    "ALLOWED_PUBLIC_KEYS": '{"keys": [{"kid": "12345", "n": "xyz"}, {"kid": "54321", "n": "xyz"}]}',  # noqa: E501
```
songofroland, line 3: let's rather do

```
('{"keys": [{"kid": "12345", "n": "xyz"}'
'{"kid": "54321", "n": "xyz"}]}')
```

tests/test_authz_authorizer.py
```diff
@@ -39,10 +39,8 @@ def test__init__(self):
         assert self.authz.permission == "permission_name"
 
     def test__repr__(self):
-        assert (
-            repr(self.authz)
-            == "Authorizer(auth_jwt=<jwt>, resource_name='test_resource', permission_name='permission_name')"
-        )
+        expected_repr = "Authorizer(auth_jwt=<jwt>, resource_name='test_resource', permission_name='permission_name')"  # noqa: E501
```
songofroland, line 42: same as above

tests/test_authz_authorizer.py
```diff
@@ -207,11 +205,6 @@ def test_get_restrictions_restricted(self):
         self.authz.denied_resource = {"city": "warszawa"}
         assert self.authz.restrictions == {"allow": ALL, "deny": {"city": "warszawa"}}
 
-    def test_sign_authz(self):
-        # noqa: E501
+    def test_sign_authz(self):  # noqa: E501
```
songofroland, line 208: What's the `noqa` for here?

tests/test_resource.py
```diff
@@ -221,6 +221,18 @@ def test_method(self):
             "request_id": ANY,
         }
 
+    def test__get_name_from_name(self):
+        class XResource(Resource):
+            _name = "test"
+
+        assert XResource.get_name() == "test"
+
+    def test__get_name_from__name__(self):
```
songofroland, line 230: it's better to keep test names `test_X_does_Y`

version
```diff
@@ -1 +1 @@
-0.3.12
```
songofroland, line 1: two versions at once?

songofroland (changes requested):

lbz/authentication.py
pdyba: re module in python is quite an expensive operation. te noqa is only because of black.

version
pdyba: there is a second PR

lbz/dev/server.py
pdyba: to enforce definition when inhariteded instead of leaving it as none.

lbz/misc.py
pdyba: that depends on usage out side LBZ.

lbz/authz.py
pdyba: 1. make sense.
2. not an issue. check below.

lbz/resource.py
pdyba: then You would need to always have it this makes it stll optional

tests/__init__.py
pdyba: unused import

lbz/authentication.py
redlickigrzegorz: I agree with @pdyba, we should use regex only if there is no other option to handle the problem. The performance may be the argument but not only - regular expressions may cause some bigger unexpected problems as in [Cloudflare](https://blog.cloudflare.com/cloudflare-outage/) in 2019 ([sekurak](https://sekurak.pl/zabilo-cale-cloudflare-przerywamy-spotkanie-wszystkie-rece-do-konsol/) was writing about that too) 😉

lbz/authz.py
redlickigrzegorz: If so, we have a bigger problem because one of the first lines tries to use request which is an instance attribute 😬 
```python
def check_permission(resource: Union[Type[Resource], Resource], permission_name: str) -> dict:
    authorization_header = resource.request.headers.get("Authorization", "")
    ...


def has_permission(resource: Union[Type[Resource], Resource], permission_name: str) -> bool:
    try:
        check_permission(resource, permission_name)
        ...
```

lbz/dev/event.py
redlickigrzegorz: Or `EVENT_TEMPLATE` 🤔  but yes, we can improve this name.

lbz/authentication.py
songofroland: Interesting, although I think the regex I proposed is really straightforward and would not be that painful.

lbz/dev/server.py
redlickigrzegorz: Probably, I understand your point of view but I think this pattern makes more sense for _"real"_ abstract classes, not for enforcing one thing in fully implemented ones. I have mixed feelings about this 🤔

lbz/misc.py
redlickigrzegorz: It was pointed in the previous [pull reqeust](https://github.com/pdyba/lambdalizator/pull/35#discussion_r672866321) - still, I think changing the type of input argument is not that good idea 🤔

lbz/authentication.py
redlickigrzegorz: They thought the same 😆  Anyway, still, the performance and probably that's not the right PR to change that.

setup.cfg
```diff
@@ -34,4 +34,12 @@ disable=
 
 [pylint.MISCELLANEOUS]
 ; http://pylint.pycqa.org/en/latest/technical_reference/features.html#miscellaneous-checker-options
-notes = FIXME
\ No newline at end of file
+notes = FIXME
+
+
+[mypy]
+ignore_missing_imports = True
```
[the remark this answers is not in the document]
redlickigrzegorz: My proposition:
1. Configuration + running `mypy setup.py lbz` (121 errors in 12 files) - the first pull request
2. Running `mypy examples tests` - the second pull request
3. `flake8` - the third pull request

The order of PRs matters because `tests` and `examples` depends on the main code - you won't fix all errors without touching the main code first 😉 `flake8` should be done separately, even if doesn't require too much work (if you want, you can do this even as a first step).

lbz/dev/server.py
pdyba: PyCharm disagrees with You, anyway isort is next on the list.

tests/fixtures/cognito_auth.py
pdyba: black come to diffrent sollution but () helped.

redlickigrzegorz (commented):
The list of 7 `mypy` errors I promised to solve:


---
> [lbz/jwt_utils.py:36:17](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/jwt_utils.py#L36): error: Returning Any from function declared to return "Dict[Any, Any]"  [no-any-return]

`mypy` is right here as it cannot detect the return type automatically from anything, so you have to help it a bit:
```python
PUBLIC_KEYS: List[dict] = []
...
        for key in PUBLIC_KEYS:  # we defined the type above, so there is no problem here
                return key
```


---
> [lbz/jwt_utils.py:57:13](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/jwt_utils.py#L57): error: Returning Any from function declared to return "Dict[Any, Any]"  [no-any-return]

Here, you need to help `mypy` as well because it is impossible to guess what `jwt.decode()` returns - you can tell directly what you expect it will be and rely on that later:
```python
decoded_jwt: dict = jwt.decode(auth_jwt_token, jwk, algorithms="RS256", audience=aud)
return decoded_jwt
```


---
> [lbz/resource.py:69:13](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/resource.py#L69): error: Returning Any from function declared to return "Response"  [no-any-return]

This code is our in 100% but it is magical and without additional type annotations even developer might have problems in guessing what is expected 😉  As always, we should write self-descriptive code:
```python
endpoint: Callable = getattr(self, self._router[self.path][self.method])
response: Response = endpoint(**self.path_params)
return response
```
_Of course, you can improve the naming here or even make a small method from this code - I wanted only to share how to resolve the problem._


---
> [lbz/dev/test.py:86:9](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/dev/test.py#L86): error: Returning Any from function declared to return "Response"  [no-any-return]
[lbz/dev/test.py:86:16](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/dev/test.py#L86): error: Too many arguments for "__call__" of "Resource"  [call-arg]
[lbz/dev/test.py:86:16](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/dev/test.py#L86): error: "Response" not callable  [operator]

All three problems are related to exactly the same line of code and the wrongly declared type:
```python
class Client:
    def __init__(self, resource: Type[Resource]):  # here, you are expecting to get a class, not its instance
        self.resource = resource
```


---
> [lbz/dev/server.py:104:35](https://github.com/pdyba/lambdalizator/blob/mypy/lbz/dev/server.py#L104): error: Argument "resource_path" to "Event" has incompatible type "Optional[str]"; expected "str"  [arg-type]

That one is a little bit tricky but `mypy` is totally right anyway. That is my proposition for this problem but it would be good to confirm it before implementing:
```python
if route is None:  # here, you inform that if the route is None you want to do something else (error)
    self._error(666, "Path not Found")
    return  # without ending processing here, the request will go further
resource = self.cls(  # pylint: disable=not-callable
    Event(
        resource_path=route,  # if the processing is not ended before, an "empty route" can be passed here
```

tests/__init__.py
redlickigrzegorz: So, I think we shouldn't import them here at all 😉

lbz/dev/test.py
```diff
@@ -11,25 +15,68 @@ class Client:
...
     """
 
-    def __init__(self, resource):
+    def __init__(self, resource: Resource):
         self.resource = resource
 
-    def post(self, resource: str, params=None, query_params=None, body=None, headers=None):
+    def post(
+        self,
+        resource: str,
```
redlickigrzegorz, line 23: The same name: `resource` but differently used within one class 😕

lbz/misc.py
```diff
@@ -16,7 +17,7 @@ class NestedDict(dict):
     Endless nested dict.
     """
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: str) -> Union[Any, dict]:
```
redlickigrzegorz, line 20: It should return just `-> Any` as `dict` is one of `Any` 😉  This class is also very generic, so returning `Any` here makes perfect sense.

lbz/misc.py
```diff
@@ -31,8 +32,8 @@ class MyClass(metaclass=Singleton):
 
     _instances: dict = {}
 
-    def __call__(cls, *args, **kwargs):
-        def _del(cls):
+    def __call__(cls, *args: Any, **kwargs: Any) -> Any:
+        def _del(cls: Any) -> None:
```
redlickigrzegorz, line 36: `cls` is defined two times, you should change the name of this argument here.

lbz/misc.py
```diff
@@ -48,34 +49,34 @@ class MultiDict(MutableMapping):
     Advanced Multi Dictionary.
     """
 
-    def __init__(self, mapping: dict):
+    def __init__(self, mapping: Union[dict, None]):
```
redlickigrzegorz, line 52: For a union like this, there is a special type: `Optional[dict]`.

lbz/misc.py
```diff
@@ -95,13 +96,13 @@ def get_logger(name: str):
 logger = get_logger(__name__)
 
 
-def error_catcher(function, default_return=False):
+def error_catcher(function: Callable, default_return: Any = False) -> Any:
```
redlickigrzegorz, line 99: That is a decorator, it returns `Callable`.

lbz/request.py
```diff
@@ -18,7 +18,7 @@
 class Request:
     """Represents request from API gateway."""
 
-    _json_body = None
+    _json_body: dict = {}
```
redlickigrzegorz, line 21: You shouldn't assign an empty dict here. That is the class attribute, so it will remain as long as the code is running. Consider this example:
```python
>>> class Request:
...     _json_body = {}
...     
...     def json_body(self):
...         return self._json_body
... 
...     def do_smth(self):
...         self._json_body["test"] = "test"
>>> a = Request()
>>> a.json_body()
{}
>>> a.do_smth()
>>> a.json_body()
{'test': 'test'}
>>> b = Request()
>>> b.json_body()
{'test': 'test'}
```

redlickigrzegorz: In general, together with `_raw_body`, it should be declared inside `__init__`.

lbz/request.py
```diff
@@ -33,21 +33,21 @@ def __init__(
         query_params: dict = None,
         user: User = None,
     ):
-        self.query_params = MultiDict(query_params)
-        self.headers = headers
-        self.uri_params = uri_params
-        self.method = method
-        self.context = context
+        self.query_params: MultiDict = MultiDict(query_params or {})
```
redlickigrzegorz, line 36: You don't have to set type if you are initializing this variable. It is obvious it will be `MultiDict` 😉

lbz/request.py
```diff
@@ -33,21 +33,21 @@ def __init__(
...
+        self.uri_params: dict = uri_params
+        self.method: str = method
+        self.context: dict = context
         self.stage_vars = stage_vars
-        self.user = user
-        self._is_base64_encoded = is_base64_encoded
-        self._body = body
+        self.user: Optional[User] = user
+        self._is_base64_encoded: bool = is_base64_encoded
+        self._body: Union[str, bytes] = body
```
redlickigrzegorz, line 44: In general, all of those attributes rely on the parameters given into `__init__`, so you don't have to specify types here. But if something is incompatible it should be changed there (as a type of input value), not here 😉

lbz/request.py
redlickigrzegorz: Also, the previous formatting was prettier 😉

lbz/resource.py
```diff
@@ -134,13 +144,14 @@ def __call__(self) -> Response:
             resp.headers.update(self.resp_headers())
         return resp
 
-    def _get_allowed_origins(self, origins: list) -> str:
+    def _get_allowed_origins(self, origins: Union[List[str], str]) -> str:
```
redlickigrzegorz, line 147: Are you sure it can be a string? 🤔  Based on the following code, this is used as a list only.

lbz/router.py
```diff
@@ -13,45 +14,45 @@ class Router(metaclass=Singleton):
...
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self._routes)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator:
         return self._routes.__iter__()
 
-    def add_route(self, routes, method, handler):
+    def add_route(self, routes: str, method: str, handler: str) -> None:
```
redlickigrzegorz, line 38: Shouldn't it be `route` (singular form) as it is a string? 🤔

lbz/router.py
```diff
@@ -13,45 +14,45 @@ class Router(metaclass=Singleton):
...
         """
         Registers handler to route and method.
         """
         self._routes[routes][method] = handler
 
 
-def add_route(route, method="GET"):
+def add_route(route: str, method: str = "GET") -> Any:
```
redlickigrzegorz, line 45: Ditto: `-> Callable`

lbz/router.py
```diff
@@ -13,45 +14,45 @@ class Router(metaclass=Singleton):
...
     """
     Flask-like wrapper for adding routes.
     """
 
-    def wrapper(func):
+    def wrapper(func: Callable) -> Any:
```
redlickigrzegorz, line 50: Ditto: `-> Callable`

tests/fixtures/cognito_auth.py
```diff
@@ -1,5 +1,7 @@
 env_mock = {
     "COGNITO_AUTHENTICATION": "1",
-    "ALLOWED_PUBLIC_KEYS": '{"keys": [{"kid": "12345", "n": "xyz"}, {"kid": "54321", "n": "xyz"}]}',
+    "ALLOWED_PUBLIC_KEYS": (
+        '{"keys": [{"kid": "12345", "n": "xyz"},' '{"kid": "54321", "n": "xyz"}]}'
```
redlickigrzegorz, line 4: You have here two strings in one line, merge them.

tests/fixtures/rsa_pair.py
```diff
@@ -1,3 +1,5 @@
+# flake8: noqa
+EXPECTED_TOKEN = "eyJhbGciOiJSUzI1NiIsImtpZCI6Ijk0OTRhZDc1LTNmNTQtNDE1NS04NGZhLWMxYTE3ZGEyMmIzNSIsInR5cCI6IkpXVCJ9.eyJhbGxvdyI6eyIqIjoiKiJ9LCJkZW55Ijp7fX0.nDqCxO2Q1iXpxzbH7syxuyqw7kCY0sDfi9RX-VSUMTRN5aWTLt1bcPw4oN_jx89-YHBzDwnwBc07RsMgpFuo4zz2LU9PF0ciYxMNX-atTNsaIn05NkXT08au2AYb0DRCDS76MZ4QNi-4mRpLrj1SD4mSCwGtc2WNw9f0J0Vm4ZCYPVW6BqpcHcaFXzcFZ6EIoooaK6GvdTOjy498lWsAXjAen2U6Jles_BwFjqW1lW_ky4WV4J9NnK3v5wWKgR1Pg4R4LpnhIXe0dU_l64JHoJA3YcYxl-qilHfoBduc3La4kRKk7FAQDIqbOv4uN03BIoDXLH5t2uJ1Sm79Pe0ngGd5pSBmfUDKOGsHtx_3_9ZKfp-E2IVS0C7r36p4Ue0gKQzn0pXxa591bxm_puJAQ399SdbmlOJsM2cVFYAtlUQvWgErc57WcUJ0Qe4jEycury7hagNbP2fLn-7Gg4gZHiZ_Ul7L6GukbDfCHnhxSS4P3t3cVtWuslZi16hDhNbOTKD95y7PXvHePvI57ALV2v0RecQ5Blwurt1OuDRSjCYXyO6U4Y9MBHcd1wMtDoVW0jjvjXvqkEhuB52Zajh_yTNnJo0OAHpuK5wldVpECGFVx1rkW1ypKqlukGIgD--m6ElKnl6jw5VWSbdh2TJsZHnzjovbQUeqZOeMxwX6SE8"
 sample_private_key = {
```
redlickigrzegorz, line 3: You should make the rest of the variables uppercase as you did with `EXPECTED_TOKEN`.

tests/test_dev_server.py
```diff
@@ -51,7 +51,7 @@ def test_my_lambda_dev_handler():
 
 
 def test_my_dev_server():
-    dev_serv = MyDevServer()
+    dev_serv = MyDevServer(MyLambdaDevHandlerHelloWorld)
```
redlickigrzegorz, line 54: `mypy` is not working for tests yet, but you provided a wrong class here 😉

tests/test_request.py
```diff
@@ -126,11 +126,8 @@ def test_accessing_user_attributes(self):
             assert isinstance(getattr(self.resp.user, letter), str)
 
     def test_headers_are_case_insensitive(self):
-        assert (
-            self.resp.headers["content-type"]
-            == self.resp.headers["CoNtEnT-TyPe"]
-            == "application/json"
-        )
+        assert self.resp.headers["content-type"] == self.resp.headers["CoNtEnT-TyPe"]
```
redlickigrzegorz, line 129: I would write it more direct instead of creating that kind of dependency:
```python
assert self.resp.headers["content-type"] == "application/json"
assert self.resp.headers["CoNtEnT-TyPe"] == "application/json"
```

tests/test_resource.py
```diff
@@ -221,6 +221,18 @@ def test_method(self):
             "request_id": ANY,
         }
 
+    def test__get_name_from__name_parameter(self):
+        class XResource(Resource):
+            _name = "test"
+
+        assert XResource.get_name() == "test"
+
+    def test__get_name_from___name___parameter(self):
```
redlickigrzegorz, line 230: Too many double underscores.

tests/test_resource.py
```diff
@@ -234,7 +246,7 @@ def teardown_method(self):
         del environ["CORS_ORIGIN"]
 
     def make_cors_handler(self, origins: List[str] = None, req_origin: str = None) -> CORSResource:
-        an_event = defaultdict(MagicMock())
+        an_event: defaultdict = defaultdict(MagicMock())
```
redlickigrzegorz, line 249: This type should be obvious as well - you don't have to annotate it that directly.

setup.cfg
```diff
@@ -34,4 +34,26 @@ disable=
...
+warn_return_any = True
+warn_unreachable = True
+warn_unused_ignores = True
+
+show_column_numbers = True
+show_error_codes = True
+
+[flake8]
+max-line-length = 99
+ignore = BLK100
```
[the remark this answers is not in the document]
pdyba: agreed but moved out 
max-complexity = 5
to next PR with TODO Comment

tests/test_resource.py
pdyba: the parametere is __name__ so in test we should add one more on both ends - that was my logic but 2 still should be sufficient

setup.cfg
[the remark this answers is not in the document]
redlickigrzegorz: That's why I initially proposed to do flake8-related things separately (before or after mypy) 😉 

As I see, there are only 8 functions/methods that are too complex, an alternative is to disable it locally, for example:
```python
def json_body(self) -> Optional[dict]:  # noqa TODO: make it less complex
```

tests/test_resource.py
redlickigrzegorz: Ok, now I get it but underscores in the name of the test means separation, so doing it this way, you decrease the readability 😉  What would you say for simple:
```python
def test__get_name__uses_class_name_attribute_when_custom_name_is_not_defined(self):
```

That is compatible with the pattern `test__tested_method__what_is_tested()` 😉

lbz/dev/server.py
```diff
@@ -94,7 +98,7 @@ def handle_request(self):
             query_params = urllib.parse.parse_qs(parsed_url.query, keep_blank_values=True)
             route, params = self._get_route_params(self.path)
             if route is None:
-                return self._error(666, "Path not Found")
+                self._error(666, "Path not Found")
```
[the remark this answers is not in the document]
pdyba: actually that method would never be used at all - it will throw duffrent error befor reaching here - fixed in upcoming PR.

lbz/dev/server.py
```diff
@@ -107,42 +111,33 @@ def handle_request(self):
...
             code = response.status_code
-            if isinstance(response, Response):
-                response = response.to_dict()
-                resp_headers = response.get("headers", {})
-                if body := response.get("body"):
-                    response = json.loads(body)
+            response_as_dict = response.to_dict()
+            resp_headers = response_as_dict.get("headers", {})
+            if body := response_as_dict.get("body"):
+                response_as_dict = json.loads(body)
```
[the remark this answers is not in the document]
pdyba: improved in next PR

lbz/dev/test.py
pdyba: renamed to path, but there is slight chance it will break stuff in other places if someone used it as kwargs.

setup.cfg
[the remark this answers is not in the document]
pdyba: yes but lets do it in seprate PR as agreed.

lbz/dev/test.py
redlickigrzegorz: Yes, that's true, but for me, it's a really small change (easy to apply) and the benefits of cleaner code are huge 😌  Thanks!

lbz/dev/test.py
```diff
@@ -11,25 +15,70 @@ class Client:
     Client created for testing purposes.
     """
 
-    def __init__(self, resource):
+    def __init__(
+        self, resource: Type[Resource]
+    ):  # here, you are expecting to get a class, not its instance
```
redlickigrzegorz, line 20: I put this comment in the proposed snippet of code only to simplify understanding, please, remove it from the code 😉

lbz/authz.py
```diff
@@ -2,10 +2,12 @@
 """
 Authorization module.
 """
+# TODO: standardize the naming convention
+# In some places, we are using permission and in other resource but it looks like sometimes they mean the same
```
redlickigrzegorz, line 6: Keep the limit of signs and also, it would be good to put this TODO closer to the place where the problem is visible:
```python
def _check_resource(self, resource: Union[dict, str]) -> None:
    # TODO: standardize the naming convention (resource != permission)
    self._deny_if_all(resource)
```

CHANGELOG.md
```diff
@@ -138,4 +138,16 @@ Released 2021-07-08
...
 
-- Add pylint, with configuration at 10.00/10.00 score
\ No newline at end of file
+- Add pylint, with configuration at 10.00/10.00 score
+
+
+### Version 0.3.13
+Released 2021-08-01
+
+- Reworks Exceptions and adds missing ones to be compatabile with all HTTP error codes.
```
redlickigrzegorz, line 147: Typo: _"compatible"_

CHANGELOG.md
```diff
@@ -138,4 +138,16 @@ Released 2021-07-08
...
+
+
+### Version 0.3.14
+Released 2021-07-30
+
+- Code quality improvments based on mypy type checking
```
redlickigrzegorz, line 153: Typo: _"improvements"_

CHANGELOG.md
```diff
@@ -138,4 +138,16 @@ Released 2021-07-08
...
+
+
+### Version 0.3.13
+Released 2021-08-01
+
+- Reworks Exceptions and adds missing ones to be compatabile with all HTTP error codes.
+
+
+### Version 0.3.14
+Released 2021-07-30
```
redlickigrzegorz, line 151: This date should be updated before the final merge 🙂

tests/test_authz_authorizer.py
redlickigrzegorz: It looks like you can eliminate every `noqa: E501` once you disabled it globally 😉

lbz/dev/server.py
pdyba: There is no suggestion on how to proceed ?

lbz/misc.py
pdyba: in lbz/resource.py:211 we are using it and passing MutliDict

lbz/dev/server.py
redlickigrzegorz: The suggestion would be to not do things like this at all, at least not in the pull request which is responsible for adding types 🙂  We can improve this part of the code (including the naming convention) in the near future if really needed.

lbz/misc.py
redlickigrzegorz: Ok, interesting, so, according to the original comment, it should be more generic: `data: MutableMapping` 😉 

And, it would be nice to point it more directly in the docstring that the return value is a simple dict, nothing else.

redlickigrzegorz: Also, the code in method you pointed looks buggy 🤔 
```python
>>> urlencode({"city": ["Berlin", "Paris"]})
'city=%5B%27Berlin%27%2C+%27Paris%27%5D'
>>> urlencode({"city": "Berlin", "city": "Paris"})
'city=Paris'
```
Correct me if I am wrong but I would say, we need to fix that sooner than later.

lbz/dev/server.py
redlickigrzegorz: [comment #685879018](https://github.com/pdyba/lambdalizator/pull/37#discussion_r685879018)
> The suggestion would be to not do things like this at all, at least not in the pull request which is responsible for adding types 🙂 We can improve this part of the code (including the naming convention) in the near future if really needed.

_Bumped, just to not forget about that._

lbz/dev/server.py
```diff
@@ -5,22 +5,26 @@
...
 
-class MyLambdaDevHandler(BaseHTTPRequestHandler):
+class MyLambdaDevHandler(BaseHTTPRequestHandler, metaclass=ABCMeta):
     """
     Mimics AWS Lambda behavior.
     """
 
-    cls = None
-    done = False
+    done: bool = False
```
redlickigrzegorz, line 22: I wrote it in a different place but in general, having class attributes that are changed during the life of an instance is risky - however, it can be improved in the future.

lbz/dev/test.py
```diff
@@ -11,25 +15,68 @@ class Client:
...
+        headers: dict = None,
+    ) -> Response:
+        return self._process(path, "DELETE", params, query_params, body, headers)
 
-    def _process(self, path, method, params, query_params, body, headers):
+    def _process(
+        self,
+        path: str,
+        method: str,
+        params: Union[dict, None],
```
redlickigrzegorz, line 75: `Optional[dict]` looks prettier 😉

lbz/misc.py
```diff
@@ -16,7 +17,7 @@ class NestedDict(dict):
     Endless nested dict.
     """
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: str) -> Any:
```
redlickigrzegorz, line 20: Shouldn't it be `key: Hashable`?

lbz/misc.py
```diff
@@ -48,34 +49,34 @@ class MultiDict(MutableMapping):
...
 
-    def __init__(self, mapping: dict):
+    def __init__(self, mapping: Optional[dict]):
         if mapping is None:
             mapping = {}
 
         self._dict = mapping
 
-    def __getitem__(self, k):
+    def __getitem__(self, k: str) -> Any:
```
redlickigrzegorz, line 58: Ditto: `k: Hashable`?

lbz/misc.py
```diff
@@ -114,7 +115,7 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
-def copy_without_keys(data: dict, *keys) -> dict:
+def copy_without_keys(data: Union[dict, MultiDict], *keys: str) -> dict:
```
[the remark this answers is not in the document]
redlickigrzegorz: I know you will fix it in the near future but right now, it will be more correct to write it this way:
```python
def copy_without_keys(data: MutableMapping, *keys: str) -> dict:
```

lbz/request.py
```diff
@@ -42,12 +39,14 @@ def __init__(
         self.user = user
         self._is_base64_encoded = is_base64_encoded
         self._body = body
+        self._json_body: Optional[dict] = None
+        self._raw_body: bytes = b""
```
redlickigrzegorz, line 43: I think it should be also equal to `None` at the start, thanks to that, you can compare to this value:
```python
if self._raw_body is None and self._body is not None:
```
In the case of an empty body, it won't be repeated 😉

lbz/request.py
```diff
@@ -69,14 +68,13 @@ def json_body(self) -> Optional[dict]:
         if content_type is None:
             return None
         if content_type.startswith("application/json"):
-            if isinstance(self._body, dict):
-                return self._body
-            if self._json_body is None:
+            if not self._json_body:
```
redlickigrzegorz, line 71: I think comparing to `None` makes more sense (previous version), because it can be set to an empty dict - thanks to that this operation won't be repeated every time when someone asks for `request.json_body()` 😉

lbz/misc.py
pdyba: Fixed in another PR.

tests/__init__.py
pdyba: nope, we should - this simplifies imports.

lbz/dev/server.py
pdyba: actually - it makes sense to add here as then it would be even more problematic for mypy
as we would have to use Union or Optional 
cls: Type(Resource) = None - wont allow us to progress.

tests/test_resource.py
```diff
@@ -310,12 +324,13 @@ def setup_method(self, _test_method, _init_mock: MagicMock) -> None:
         self.resource = PaginatedCORSResource({}, [])
         self.resource.path = "/test/path"
         self.resource.urn = "/test/path"
-        self.resource.request = SimpleNamespace(
-            query_params={
-                "test": "param",
-                "another": "example",
+        req.query_params = MultiDict(
```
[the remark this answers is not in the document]
pdyba: It was being already changed that either way. Actually using SimpleNamespace was wrong from the beggining.

lbz/dev/misc.py
```diff
@@ -69,16 +19,16 @@ def __init__(
         self,
         resource_path: str,
         method: str,
-        body=None,
-        query_params=None,
-        path_params=None,
-        headers=None,
-    ):
-        super().__init__(**json.loads(EVENT))
+        body: dict = None,
```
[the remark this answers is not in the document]
pdyba: We json load it probably befohand but yes that could be improved.

tests/__init__.py
redlickigrzegorz: That is true but all _"fixtures"_ should be gathered in `tests/fixtures/__init__.py`, not here 😉

lbz/dev/server.py
redlickigrzegorz: You can always do: `cls: Type[Resource] = None  # type: ignore` 😉  and it would be the easiest thing in this case.

But ok, let's say, we will improve it in the future, just to simplify the code.

tests/test_resource.py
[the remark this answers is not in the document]
redlickigrzegorz: `SimpleNamespace` is wrong when we talk about the type in general but if we need a simple object that contains the `query_params` attribute, it is a pretty nice simplification 😉 

I didn't see the wrong usage of `req` but generally, defining it as a global variable is a wrong idea, so it would be good to change that sooner than later (according to the TODO, instead of that, we should have a fixture) - can be done in the near future.

redlickigrzegorz (approved):

PR #38, opened 2021-08-18
Query params bug fix

pdyba:


lbz/misc.py
```diff
@@ -85,34 +114,20 @@ def getlist(self, k: Hashable) -> list:
...
-    return logger_obj
-
-
-logger = get_logger(__name__)
-
+    def items(self) -> Iterator:  # type: ignore[override]
+        # override as Mapping.items method is expected to be returning  set and set looses order.
+        for key, values in self._dict.items():
+            for value in values:
+                yield key, value
```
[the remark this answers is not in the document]
redlickigrzegorz: In general, doing anything with this method breaks the key functionality of the `MultiDict` class - I think we should talk for 15-30 minutes to gather information of what we want to achieve 🙂

lbz/resource.py
```diff
@@ -207,6 +207,8 @@ def get_pagination(self, total_items: int, limit: int, offset: int) -> dict:
 
     @property
     def _pagination_uri(self) -> str:
-        if query_params := copy_without_keys(self.request.query_params, "offset", "limit"):
-            return f"{self.urn}?{urlencode(query_params)}&offset={{offset}}&limit={{limit}}"
+        if query_params := self.request.query_params.clean_keys(["offset", "limit"]):
+            # TODO: Make MultiDict be a Mapping not a object to remove type ignore below
+            encoded_params = urlencode(query_params)  # type: ignore
```
redlickigrzegorz, line 212: When you improve the code around `MultiDict.items()`, you should start using the `doseq=True` argument to put each value separately to the URL 🙂

lbz/resource.py
```diff
@@ -207,6 +207,8 @@ def get_pagination(self, total_items: int, limit: int, offset: int) -> dict:
 
     @property
     def _pagination_uri(self) -> str:
-        if query_params := copy_without_keys(self.request.query_params, "offset", "limit"):
-            return f"{self.urn}?{urlencode(query_params)}&offset={{offset}}&limit={{limit}}"
+        if query_params := self.request.query_params.clean_keys(["offset", "limit"]):
```
redlickigrzegorz, line 210: Removing these keys to achieve URL encoding makes sense but you cannot remove them permanently from the original `query_params` because they will be needed. I mean a situation like this:
```python
self.request.query_params["limit"]  # returns the limit provided from the outside
self._pagination_uri()  # returns URI but removes offset and limit
self.request.query_params["limit"]  # raises KeyError
```

You need to create a copy of those query parameters just to handle this case. Also, it would be good to cover it by the test, just to not break it in the future 😉

tests/test_resource.py
```diff
@@ -346,6 +346,25 @@ def test_get_pagination(self):
...
+    def test_get_pagination_multifield_query_params(self):
+        expected_prefix = "/test/path?test=param&another=example&another=example2"
+        req.query_params = MultiDict(
+            {
+                "test": ["param"],
+                "another": ["example", "example2"],
+            }
+        )
+        self.resource.request = req
+        assert self.resource.get_pagination(total_items=100, offset=20, limit=10) == {
```
redlickigrzegorz, line 358: Please, place an empty line before this one just to visually separate steps of this test 😉

lbz/misc.py
```diff
@@ -85,34 +114,20 @@ def getlist(self, k: Hashable) -> list:
...
-            return default_return
-
-    return wrapped
+    def clean_keys(self, keys: List[Hashable]) -> object:
+        for key in keys:
+            self.safe_delete(key)
+        return self
 
 
 def copy_without_keys(data: MutableMapping, *keys: str) -> dict:
```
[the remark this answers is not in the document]
pdyba: backwards compatibility - we can remove it only in 0.4+

lbz/misc.py
```diff
@@ -85,34 +114,20 @@ def getlist(self, k: Hashable) -> list:
...
-            return function(*args, **kwargs)
-        except Exception as error:  # pylint: disable=broad-except
-            if len(args) > 0 and hasattr(args[0], "logger"):
-                args[0].logger.exception(error)
-            else:
-                logger.exception(error)
-            return default_return
-
-    return wrapped
+    def clean_keys(self, keys: List[Hashable]) -> object:
```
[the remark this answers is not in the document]
pdyba: from __future__ import annotations -> nice one 
added comment why it cannot return none.

lbz/misc.py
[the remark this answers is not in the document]
pdyba: added commnet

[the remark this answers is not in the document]
redlickigrzegorz: Fair point, it doesn't harm to keep it either 👍

lbz/misc.py
```diff
@@ -113,10 +126,3 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
             return default_return
 
     return wrapped
-
-
-def copy_without_keys(data: MutableMapping, *keys: str) -> dict:
```
redlickigrzegorz, line 118: I don't see any value in changing the functions' order here. So if not needed, it would be good to not touch the code at all - just a tip for the future on how to improve reviewers life 😉

lbz/misc.py
```diff
@@ -2,12 +2,14 @@
 """
 Misc Helpers of Lambda Framework.
 """
+from __future__ import annotations
```
redlickigrzegorz, line 5: You don't need that in the final implementation.

setup.py
```diff
@@ -1,9 +1,9 @@
 from setuptools import setup
 
-with open("version") as file:
+with open("version", encoding="UTF-8") as file:
```
redlickigrzegorz, line 3: You can use also the simpler version which we have in other repository:
```python
pathlib.Path("version").read_text("utf-8")
```

tests/test_request.py
```diff
@@ -51,12 +50,11 @@ def setup_method(self):
...
         self.id_token = encode_token(self.cognito_user)
         self.resp = Request(
-            headers=CIMultiDict({"Content-Type": "application/json"}),
-            uri_params={},
-            # pylint issue #214
             method="GET",
+            uri_params={},
             body="",
             context={},
+            headers=CIMultiDict({"Content-Type": "application/json"}),  # pylint issue #214
```
redlickigrzegorz, line 57: It looks like with the new version of `pylint`, [the configuration](http://pylint.pycqa.org/en/latest/technical_reference/features.html#similarities-checker) started to work correctly. That means we can increase the number of lines taken for similarity comparison:
```cfg
# setup.cfg

[pylint.SIMILARITIES]
min-similarity-lines = 9
```
I don't want to say we should set `9` but it is the minimal number with which we don't get an error. Also, in the future, we will have to do something with that anyway 😉

lbz/resource.py
```diff
@@ -207,6 +207,8 @@ def get_pagination(self, total_items: int, limit: int, offset: int) -> dict:
 
     @property
     def _pagination_uri(self) -> str:
-        if query_params := copy_without_keys(self.request.query_params, "offset", "limit"):
-            return f"{self.urn}?{urlencode(query_params)}&offset={{offset}}&limit={{limit}}"
+        if query_params := self.request.query_params.original_items("offset", "limit"):
```
redlickigrzegorz, line 210: When I see it right now, I need to say that it is misleading 😬  I think, much more understandable would be to use keyword argument here:
```python
def original_items(self, keys_to_skip: Iterable[Hashable] = None) -> List[tuple]:
    keys_to_skip = keys_to_skip or []
    ...

self.request.query_params.original_items(keys_to_skip=["offset", "limit"])
```

lbz/resource.py
```diff
@@ -207,6 +207,8 @@ def get_pagination(self, total_items: int, limit: int, offset: int) -> dict:
...
+            # TODO: Make MultiDict be a Mapping not a object to remove type ignore below
```
redlickigrzegorz, line 211: This TODO is no longer valid 😉  But to satisfy `mypy` you can change the type of `query_params`:
```python
def original_items(...) -> Sequence[Tuple[Any, Any]]:
    ...
```

redlickigrzegorz (approved):

tests/test_request.py
pdyba: I wouldn't put that number at all - this is valid thing for even 1 line, but in this case does not make sense.

redlickigrzegorz: Yep, `pylint` messed up with those similarities 😕  Let's keep workarounds.

PR #39, opened 2021-08-18
Authz meta

pdyba:


lbz/authz/__init__.py
```diff
@@ -0,0 +1,5 @@
+# compatibility with previous versions
```
redlickigrzegorz, line 1: You don't have to write it directly - that is a common pattern to put imports like those into the `__init__.py`module.

lbz/authz/collector.py
```diff
@@ -0,0 +1,16 @@
+from lbz.misc import Singleton
+
+
+class AuthCollector(metaclass=Singleton):
+    # TODO: Add uni tests
+    def __init__(self) -> None:
+        self.possible_authz: list = []
```
redlickigrzegorz, line 7: Maybe it could be named `self.possible_permissions`? We should start being consistent in naming to simplify the code and its results.

lbz/authz/collector.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def add_authz(self, permission: str) -> None:
+        self.possible_authz.append(permission)
```
redlickigrzegorz, line 10: The dump method bases on the resource but here you are adding all permissions into one list, I think that is not fully correct 🤔

lbz/authz/collector.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def dump_authz(self, resource: str, guest_permissions: dict) -> dict:
+        return {resource: self.possible_authz, "guest_permissions": guest_permissions}
```
redlickigrzegorz, line 13: If we want to have a resource key, everything should be under this key, not only a list of permissions. But I would suggest keeping it simple:
```python
{
    "possible_permissions": {},
    "guest_permisssions": {},
}
```

lbz/authz/collector.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+authz_collector = AuthCollector()
```
redlickigrzegorz, line 16: The naming should be consistent here too. The class should be named `AuthzCollector` as it is inside the `authz` package 😉

lbz/authz/wrappers.py
```diff
@@ -0,0 +1,64 @@
+# coding=utf-8
+"""
+Authorization module.
```
redlickigrzegorz, line 3: You named this module `wrappers.py` so here, you should update that as well.

redlickigrzegorz: But in general, I am not sure if this name is fully correct. Only one function here is a decorator. Maybe we should put this everything into the `permissions.py` module together with the `AuthzCollector` class 🤔

What do you think?

lbz/authz/wrappers.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import Callable, Union, Type, Any
+
+from lbz.exceptions import PermissionDenied, Unauthorized
+from lbz.resource import Resource
+from lbz.authz.collector import authz_collector
+from lbz.authz.authorizer import Authorizer
+
+
+def check_permission(resource: Union[Type[Resource], Resource], permission_name: str) -> dict:
```
redlickigrzegorz, line 15: You changed the type of resource to the wrong one. It should be equal to `resource: Resource`.

lbz/resource.py
```diff
@@ -111,6 +112,11 @@ def get_guest_authorization() -> dict:
         Place to configure default authorization. That will be used when Authorization Header is not in place.
         """
 
+    def get_all_possible_authz(self) -> dict:
+        return authz_collector.dump_authz(
```
redlickigrzegorz, line 116: Wouldn't it be simpler to keep permissions closer to the resource? I mean something similar to `_router = Router()` that we already have 🤔

lbz/authz/wrappers.py
```diff
@@ -0,0 +1,64 @@
...
+    return True
+
+
+def authorization(permission_name: str = None) -> Callable:
+    """
+    Wrapper for easy adding authorization requirement.
+    """
+
+    def decorator(func: Callable) -> Callable:
+        authz_collector.add_authz(permission_name or func.__name__)
```
redlickigrzegorz, line 55: Based on one of the previous comments: you should store the information about permissions based on the resource. Otherwise, all will be put into one big list - we will lose the possibility to say directly which permissions are related to which resource.

lbz/authz/wrappers.py
pdyba: Its to avoid import issues - circular imports. - fixed description

lbz/authz/wrappers.py
pdyba: added relevant comment

lbz/authz/wrappers.py
```diff
@@ -0,0 +1,70 @@
...
+    Does not raise.
+    """
+    try:
+        check_permission(resource, permission_name)
+    except (Unauthorized, PermissionDenied):
+        return False
+    return True
+
+
+authz_collector = AuthzCollector()
```
redlickigrzegorz, line 48: Why not initialize it inside the `collector.py`? It sounds reasonable to have it there instead of keeping it here between functions 🤔

lbz/authz/collector.py
```diff
@@ -0,0 +1,28 @@
+from lbz.misc import Singleton
+
+
+class AuthzCollector(metaclass=Singleton):
+    # TODO: Add uni tests
```
redlickigrzegorz, line 5: This TODO is not needed once you created unit tests 😉

lbz/authz/collector.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(
+        self,
```
redlickigrzegorz, line 7: Place it in one line 😉

lbz/authz/collector.py
```diff
@@ -0,0 +1,28 @@
...
+        self.guest_permissions: dict = {}
+
+    def set_resource(self, resource_name: str) -> None:
+        self.resource = resource_name
+
+    def set_guest_permissions(self, guest_permissions: dict) -> None:
+        self.guest_permissions = guest_permissions
+
+    def add_authz(self, permission: str) -> None:
+        self.possible_permissions.append(permission)
```
redlickigrzegorz, line 20: In the near future, we will want to add more information about the permission. Maybe it is worth making it a dictionary right now?
```python
self.possible_permissions[permission] = None 
```
Currently `None` but in the future a description or more complicated structure. What do you think?

lbz/authz/collector.py
```diff
@@ -0,0 +1,28 @@
...
+        self.resource = ""
+        self.guest_permissions: dict = {}
+
+    def set_resource(self, resource_name: str) -> None:
+        self.resource = resource_name
+
+    def set_guest_permissions(self, guest_permissions: dict) -> None:
+        self.guest_permissions = guest_permissions
+
+    def add_authz(self, permission: str) -> None:
```
redlickigrzegorz, line 19: `permission_name` just to stay consistent.

lbz/authz/collector.py
```diff
@@ -0,0 +1,28 @@
...
+    # TODO: Add uni tests
+    def __init__(
+        self,
+    ) -> None:
+        self.possible_permissions: list = []
+        self.resource = ""
+        self.guest_permissions: dict = {}
+
+    def set_resource(self, resource_name: str) -> None:
+        self.resource = resource_name
```
redlickigrzegorz, line 14: If we want to store here the name, it should be `self.resource_name` (the same for the name of the method) just to not mislead anyone that this is a full `Resource` object.

lbz/resource.py
```diff
@@ -40,6 +41,7 @@ def get_name(cls) -> str:
 
     def __init__(self, event: dict):
         self._load_configuration()
+        self._authz_collector = AuthzCollector()
```
redlickigrzegorz, line 44: Why not on the class level? e.g. right below `_router`.

lbz/resource.py
```diff
@@ -55,6 +57,8 @@ def __init__(self, event: dict):
             is_base64_encoded=event.get("isBase64Encoded", False),
             query_params=event["multiValueQueryStringParameters"],
         )
+        self._authz_collector.set_resource(self.get_name())
+        self._authz_collector.set_guest_permissions(self.get_guest_authorization() or {})
```
redlickigrzegorz, line 61: I think `get_guest_authorization()` should return an empty dict by default instead of putting here `or {}` 😉

lbz/resource.py
```diff
@@ -111,6 +115,9 @@ def get_guest_authorization() -> dict:
         Place to configure default authorization. That will be used when Authorization Header is not in place.
         """
 
+    def get_all_possible_authz(self) -> dict:
```
redlickigrzegorz, line 118: `get_authz_data()`? It returns more than only possible authz.

redlickigrzegorz (approved):

lbz/authz/wrappers.py
redlickigrzegorz: Ok, I see it but still, the name `wrappers.py` is not fully precise 🤔

PR #40, opened 2021-08-18
Tests improvements

pdyba:


lbz/resource.py
```diff
@@ -117,7 +117,8 @@ def get_guest_authorization() -> dict:
         return {}
 
     def get_authz_data(self) -> dict:
-        return self._authz_collector.dump()
+        dump: dict = self._authz_collector.dump()
```
redlickigrzegorz, line 120: That change doesn't make much sense because `.dump()` is our internal method and contains a type annotation - the type will be discovered automatically. I don't see any other reason to create a separate variable for that but if you see, please share it with me 😉

lbz/authz/__init__.py
```diff
@@ -1,4 +0,0 @@
-# flake8: noqa
```
redlickigrzegorz, line 1: You cannot just remove `__init__.py` module. That makes `authz` a normal directory (it should be a Python package) and probably, it won't be found during building the release for `lbz` (run `python setup.py sdist` to see that).

Please, describe here the problem you wanted to avoid, maybe there is a better idea to solve that 😉

lbz/authz/authorizer.py
```diff
@@ -132,8 +132,7 @@ def _check_allow_and_set_resources(self) -> None:
         ):
             return
         if d_domain := self.allow.get(self.resource):
-            if self._allow_if_allow_all(d_domain):
-                return
+            self._allow_if_allow_all(d_domain)
```
redlickigrzegorz, line 135: I didn't check tests yet but it doesn't look like a good change. Consider the case hen `d_domain` is equal to `ALL ("*")`. In the line below, you will get `AttributeError` then 🤔

redlickigrzegorz: Oh, the `self._allow_if_allow_all` is also used above 🤔  So, correct me if I am wrong but in my opinion, here, it shouldn't be triggered at all then.

lbz/authz/authorizer.py
```diff
@@ -148,7 +147,7 @@ def restrictions(self) -> dict:
         return {"allow": self.allowed_resource, "deny": self.denied_resource}
 
     @staticmethod
-    def sign_authz(authz_data: dict, private_key_jwk: dict) -> str:
+    def sign_authz(authz_data: dict, private_key_jwk: dict) -> Union[str, Any]:
```
redlickigrzegorz, line 150: That is not true, you have here only one return statement in this method and it is declared to always return a string 😉

lbz/authz/collector.py
```diff
@@ -24,5 +27,8 @@ def dump(self) -> dict:
             }
         }
 
+    def clean(self) -> None:
+        self.possible_permissions = {}
```
redlickigrzegorz, line 31: Shouldn't it also unset the rest of the attributes? I mean `resource_name` and `guest_permissions` 😉

lbz/authz/decorators.py
```diff
@@ -29,7 +29,8 @@ def check_permission(resource: Resource, permission_name: str) -> dict:
         policy_override=guest_authorization_policy,
     )
     authorizer.check_access()
-    return authorizer.restrictions
+    restrictions: dict = authorizer.restrictions
```
redlickigrzegorz, line 32: The same thing as before, there is no point to create a separate variable named `restrictions` if we want to not use it and just return one line below. The previous version was fully correct and nicer 😉

lbz/dev/misc.py
```diff
@@ -40,3 +40,12 @@ def __init__(
...
     def __repr__(self) -> str:
         return f"<Fake Event {self['method']} @ {self['path']} body: {self['body']}>"
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, dict):
+            self_copy = self.copy()
+            self_copy.pop("requestContext")
+            other_copy = other.copy()
+            other_copy.pop("requestContext")
+            return hash(json.dumps(self_copy)) == hash(json.dumps(other_copy))
```
redlickigrzegorz, line 50: That doesn't seem to be the right way to compare dictionaries, especially that simple `dict` does not guarantee the order of elements:
```python
dict_1 = {"a": 1, "b": 2}
dict_2 = {"b": 2, "a": 1}

assert dict_1 == dict_2  # True
assert hash(json.dumps(dict_1)) == hash(json.dumps(dict_2))  # False
```

lbz/dev/server.py
```diff
@@ -24,19 +26,21 @@ class MyLambdaDevHandler(BaseHTTPRequestHandler, metaclass=ABCMeta):
     @property
     @abstractmethod
     def cls(self) -> Type[Resource]:
-        pass
+        """
+        DevServer needs to inherit a Resource.
```
redlickigrzegorz, line 30: This is a misleading sentence as the current pattern does not rely on inheriting: resource is just assigned to `cls` on the class declaration level.

In the past, I wanted to remove this `cls` abstract method at all, or at least write a proper TODO because things like this need to be done simpler (developer-friendly).

lbz/dev/server.py
```diff
@@ -56,9 +60,9 @@ def _get_route_params(self, org_path: str) -> Tuple[Union[str, None], Union[dict
                         acc += 1
                 if len(path) == acc:
                     return org_route, params
-        return None, None
+        raise ValueError
```
redlickigrzegorz, line 63: Please place here an error message to say directly what happened - it will simplify resolving problems in the future 😉

lbz/dev/server.py
```diff
@@ -56,9 +60,9 @@ def _get_route_params(self, org_path: str) -> Tuple[Union[str, None], Union[dict
...
 
-    def _send_json(self, code: int, obj: dict, headers: dict = None) -> None:
+    def _send_json(self, code: int, obj: Union[str, dict], headers: dict = None) -> None:
```
redlickigrzegorz, line 65: As I see, `obj` is only used at the end of this method, to be JSON dumped. But if it is already a string, that line will produce a different result - I am not sure if it is something that should be accepted, at least it doesn't look so:
```python
>>> json.dumps({"data": []}, indent=4, sort_keys=True)
'{\n    "data": []\n}'
>>> json.dumps('{"data": []}', indent=4, sort_keys=True)
'"{\\"data\\": []}"'
```

lbz/dev/server.py
```diff
@@ -110,15 +110,8 @@ def handle_request(self) -> None:
                     body=request_obj,
                 )
             )
-            response = resource()
-            code = response.status_code
-            response_as_dict = response.to_dict()
```
redlickigrzegorz, line 115: There is a piece of logic inside the `.to_dict()` method which makes sense to me but you skipped it and you always returns raw `body`:
```python
body = (
    json.dumps(self.body, separators=(",", ":"), default=str)
    if self.is_json
    else self.body
)
```
What do you think about that? 🤔

lbz/dev/server.py
```diff
@@ -142,7 +135,7 @@ def do_OPTIONS(self) -> None:  # pylint: disable=invalid-name
         self.handle_request()
 
 
-class MyDevServer:
+class MyDevServer(Thread):
```
redlickigrzegorz, line 138: Can you describe in a few words what gives us this inheritance? I can investigate it on my own but I thought it will be simpler/faster to get it from you 😉

Makefile
```diff
@@ -6,7 +6,16 @@ black-check:
 
 test:
 	coverage run --include 'lbz/*' -m pytest "tests"
-	coverage report --skip-covered
+	coverage report --skip-covered -m
+
+real-coverage:
+	for file in $$(find lbz -type f \( -name "*.py" -and ! -name "*_.py" -and ! -name "event.py" \)); do       \
```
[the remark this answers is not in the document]
pdyba: its my own, its not a feature its supporting script. We can skip it for now.
There orginal rule was to name tests test_module_its_testing.py.
Its not in the pipeline so we can improve that further in the future.
Spliting the test in packages makes sense but the makefile script is a PITA.

lbz/authz/__init__.py
redlickigrzegorz: I checked it more deeply and it looks like the bug has been introduced in #39. A newly created package hasn't been mentioned in the `lbz` [packages](https://github.com/pdyba/lambdalizator/blob/master/setup.py#L9) - it means v0.3.16 is broken because it does not contain `authz` at all.

But still, we cannot remove `__init__.py` from here.

Makefile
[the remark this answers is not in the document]
redlickigrzegorz: Ok, so just remove that or comment out with putting a nice TODO and place the test modules in the proper packages (we have two at this moment: `authz` and `dev`) 😉  We will return to that with a smaller PR.

tests/test_router.py
```diff
@@ -1,48 +1,49 @@
...
 
 class TestRouter:
-    def setup_method(self):
+    def setup_method(self) -> None:
         # pylint: disable= attribute-defined-outside-init
         self.router = Router()
         self.router.add_route("/", "GET", "x")
 
-    def teardown_method(self, _test_method):
+    def teardown_method(self, _test_method: Callable) -> None:
```
redlickigrzegorz, line 16: I think you could use [types.MethodType](https://docs.python.org/3/library/types.html#types.MethodType) here instead 😉 

_Side note:_ we don't have `mypy` enabled for tests, so they are not checked automatically - introducing types into this big pull request may make a review longer 😐

tests/test_response.py
```diff
@@ -67,3 +69,7 @@ def test_response_body_passes_through_when_base64(self):
             "statusCode": 666,
             "isBase64Encoded": True,
         }
+
+    def test_unsported_header_raises(self) -> None:
```
redlickigrzegorz, line 73: typo: _"un**suppo**rted"_

tests/test_response.py
```diff
@@ -67,3 +69,7 @@ def test_response_body_passes_through_when_base64(self):
...
+        with pytest.raises(RuntimeError):
+            Response(b"xxx", status_code=666)
```
redlickigrzegorz, line 75: The name of the test is suggesting that the header is tested but here you put `body` and the `status_code`. So, what exactly is tested here? 🤔

redlickigrzegorz: Also, the body expects to get `str` or `dict` but you provided `bytes` - `mypy` would catch that, I guess.

tests/test_response.py
redlickigrzegorz: It would be good to place a full sentence too, including the reason - it will be valuable for future developers.

tests/test_request.py
```diff
@@ -37,19 +34,10 @@ def test__init__(self):
 
 
 class TestRequest:
-    def setup_method(self):
+    @pytest.fixture(autouse=True)
```
redlickigrzegorz, line 37: You shouldn't mark `setup_method` as a fixture - that is a wrong usage of this functionality. You should prepare a normal fixture above which will be used in the tests:
```python
@pytest.fixture
def sample_request(user: User) -> Request:
    return Request(
        ...
        user=user,
    )
```

redlickigrzegorz: Or, if you want to keep the `setup_method` as it is right now, you should prepare a factory that can be used within this method. But it will be rather a normal function - it won't break anything in the `pytest` fixtures 😉

tests/test_misc.py
```diff
@@ -1,16 +1,19 @@
...
+
 from lbz.misc import MultiDict, NestedDict, Singleton, error_catcher, get_logger
 
 
-def test_nested_dict():
+def test_nested_dict() -> None:
     nest = NestedDict()
     nest["a"]["b"]["c"]["d"]["e"] = "z"
     assert nest == {"a": {"b": {"c": {"d": {"e": "z"}}}}}
+    assert nest["a"]["b"]["c"]["d"]["e"] == "z"
```
redlickigrzegorz, line 13: If the above is true, this line just tests that Python's `dict` is working correctly 😉 

I would say, it is not needed but it is just a thought.

tests/test_misc.py
```diff
@@ -43,7 +46,18 @@ def test_multi_dict():
     assert multi_dict.original_items() == [("b", ["abc"])]
 
 
-def test_get_logger(caplog):
+def test_multi_dict_init_empty() -> None:
+    multi_dict = MultiDict(None)
```
redlickigrzegorz, line 50: If it is a desired feature, maybe the default value for `mapping` can be equal to `None` 😉 

Thanks to that, we don't have to pass it and simple `MultiDict()` will behave the same.

tests/test_misc.py
```diff
@@ -43,7 +46,18 @@ def test_multi_dict():
...
+    assert multi_dict._dict == {}  # pylint: disable=protected-access
+
+
+def test_multi_dict_index_error() -> None:
```
redlickigrzegorz, line 54: Based on the test implementation, I think the name should be improved, something like:
```python
def test__multi_dict__raises_key_error_if_value_of_key_is_empty_list() -> None
```

On the other hand, I have to say that this is a bit counterintuitive and I would treat it as a bug but maybe I am wrong - documentation may help here a little 😉

tests/test_jwt_utils_import_error.py
```diff
@@ -0,0 +1,11 @@
+from unittest.mock import patch
+
+import pytest
+
+
+def not_working_test() -> None:  # TODO: finish me! - lbz/jwt_utils.py:26 mocking hell
```
redlickigrzegorz, line 6: Please use `@pytest.mark.skip` to skip tests that you don't want to run. It will be at least shown in the report 😉  Also, it would be good to write a more descriptive TODO, at the first look, I don't have a clue what you wanted to achieve here.

tests/test_authentication.py
```diff
@@ -8,67 +8,56 @@
...
-allowed_audiences = [str(uuid4()), str(uuid4())]
+from tests.utils import encode_token, allowed_audiences
 
 
 @patch("lbz.jwt_utils.PUBLIC_KEYS", [SAMPLE_PUBLIC_KEY])
 @patch("lbz.jwt_utils.ALLOWED_AUDIENCES", allowed_audiences)
 class TestAuthentication:
-    def setup_class(self):
-        # pylint: disable=attribute-defined-outside-init
+    @pytest.fixture(autouse=True)
```
redlickigrzegorz, line 17: _ditto:_ wrong usage of `pytest.fixture`

tests/test_authentication.py
```diff
@@ -8,67 +8,56 @@
...
+
+    def test__repr__username(self, user_username) -> None:
+        assert self.sample_user.__repr__() == f"User username={user_username}"
 
-    def test__repr__username(self):
-        username = str(uuid4())
-        sample_user = User(encode_token({"cognito:username": username}))
-        assert sample_user.__repr__() == f"User username={username}"
+        sample_user_2 = User(encode_token({"type": "x"}))
+        assert sample_user_2.__repr__() == "User"
```
redlickigrzegorz, line 31: Please put these assertions into separate tests - they represent two cases.

tests/test_authentication.py
```diff
@@ -8,67 +8,56 @@
...
+    def test_decoding_user_raises_unauthorized_when_invalid_public_key(self) -> None:
         with pytest.raises(Unauthorized), patch(
             "lbz.jwt_utils.PUBLIC_KEYS",
             [{**SAMPLE_PUBLIC_KEY.copy(), "n": str(uuid4())}],
         ):
             User(self.id_token)
 
-    def test_loading_user_parses_user_attributes(self):
-        parsed = self.cognito_user.copy()
+    def test_loading_user_parses_user_attributes(self, user_cogniot) -> None:
```
redlickigrzegorz, line 51: typo: _"user_cogni**to**"_

tests/test_resource.py
```diff
@@ -233,28 +161,90 @@ def test_method(self):
...
+
+class TestResourceAuthentication:
+    @pytest.fixture(autouse=True)
+    def setup_class(self, sample_event) -> None:
+        class XResource(Resource):
+            @add_route("/")
+            def test_method(self) -> None:
+                return Response("x")
+
+        self.res = XResource  # pylint: disable=attribute-defined-outside-init
```
redlickigrzegorz, line 201: It would be good to name it `self.resource_cls` or something similar as you assigned the whole class to this attribute 😉

tests/test_resource.py
```diff
@@ -104,123 +71,84 @@ def test_method(self):
...
+    def test_unsupported_method(self, sample_request) -> None:
+        event_wrong_method = Event(
+            resource_path="/",
+            method="" "PRINT",
+            headers={},
+            path_params={},
+            query_params={},
+            body=sample_request,
+        )
+        res = Resource(event_wrong_method)()
```
redlickigrzegorz, line 96: Didn't you mean `resp` or even `response`? You are using `res` as a shortcut for `resource` in some places - it can be misleading 😉

tests/test_resource.py
```diff
@@ -233,28 +161,90 @@ def test_method(self):
...
+    def test_unauthorized_when_no_matching_key_in_env_variable(self) -> None:
+        authentication_token = jwt.encode({"kid": "foobar"}, "")
+        resp = self.res({**self.event, "headers": {"authentication": authentication_token}})()
+        assert resp.status_code == HTTPStatus.UNAUTHORIZED
+
+    def test_unauthorized_when_jwt_header_malformed(self) -> None:
+        resp = self.res({**self.event, "headers": {"authentication": "12345"}})()
+        assert resp.status_code == HTTPStatus.UNAUTHORIZED
+
+    def test_unauthorized_when_authentication_not_configured_correctly(self) -> None:
```
redlickigrzegorz, line 233: Here you are initializing and calling the instance of the Resource class - it is hard to guess what you mean byt `not_configured_correctly` 😉  It would be very good to show that more directly or write a short comment about it.

tests/test_resource.py
```diff
@@ -233,28 +161,90 @@ def test_method(self):
...
+        with patch("lbz.resource.env", {}):
```
redlickigrzegorz, line 234: I think you should use `patch.dict` instead.

tests/test_resource.py
```diff
@@ -233,28 +161,90 @@ def test_method(self):
...
 ORIGIN_EXAMPLE = "https://api.example.com"
 
 
 class TestCORSResource:
-    def setup_method(self):
+    def setup_class(self) -> None:
         environ["CORS_ORIGIN"] = f"{ORIGIN_LOCALHOST},{ORIGIN_EXAMPLE}"
 
-    def teardown_method(self):
+    def teardown_class(self) -> None:
```
redlickigrzegorz, line 247: Both `setup_class` and `teardown_class` should be marked as class methods 😉

tests/test_resource.py
```diff
@@ -263,63 +253,60 @@ def make_cors_handler(self, origins: List[str] = None, req_origin: str = None) -
...
+            (ORIGIN_EXAMPLE, ORIGIN_EXAMPLE),
+            ("invalid_origin", ORIGIN_LOCALHOST),
+        ],
+    )
+    def test_cors_origin_headers_from_env_are_correct(self, req_origin, orig_outcome) -> None:
+        headers = self.make_cors_handler(req_origin=req_origin).resp_headers_json[
             ALLOW_ORIGIN_HEADER
         ]
-        assert headers == ORIGIN_EXAMPLE
+        assert headers == orig_outcome
```
redlickigrzegorz, line 268: Please improve naming around this test. Why would `headers` be equal to `orig_outcome`? What is `orig_outcome`? Those are questions I asked reading it for the first time 😬

tests/test_resource.py
```diff
@@ -263,63 +253,60 @@ def make_cors_handler(self, origins: List[str] = None, req_origin: str = None) -
...
-    def test_cors_origin_headers_from_wildcard_star(self):
+    def test_cors_origin_headers_from_wildcard_star(self) -> None:
         assert self.make_cors_handler(origins=["*"]).resp_headers_json[ALLOW_ORIGIN_HEADER] == "*"
         cors_handler = self.make_cors_handler(origins=["*"], req_origin="http://localhost:3000")
         assert cors_handler.resp_headers_json[ALLOW_ORIGIN_HEADER] == "*"
 
-    def test_all_headers(self):
+    def test_all_headers(self) -> None:
         content_type = "image/jpeg"
+        expected_header = "Content-Type, X-Amz-Date, Authentication, Authorization, X-Api-Key, X-Amz-Security-Token"  # noqa: E501
```
redlickigrzegorz, line 294: Previously it was placed as a multi-line string below - I liked that more 😉

tests/test_resource.py
```diff
@@ -330,6 +317,15 @@ def test_options_request(self):
             ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
         }
 
+    @patch.object(CORSResource, "resp_headers", return_value={})
+    def test_not_allow_oring_in_header(self, resp_headers_mock: MagicMock) -> None:
+        an_event = defaultdict(MagicMock())
+        an_event["method"] = "GET"
+        an_event["headers"] = {"origin": "notlocalhost"}
+        resp = CORSResource(an_event, ["GET", "POST"], origins="localhost")()
```
redlickigrzegorz, line 325: The `origins` param expects to get a `List` of strings, not a single string.

tests/test_resource.py
```diff
@@ -330,6 +317,15 @@ def test_options_request(self):
             ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
         }
 
+    @patch.object(CORSResource, "resp_headers", return_value={})
+    def test_not_allow_oring_in_header(self, resp_headers_mock: MagicMock) -> None:
+        an_event = defaultdict(MagicMock())
```
redlickigrzegorz, line 322: Why didn't you use `self.make_cors_handler`? It looks like it builds exactly the same thing as you implemented here 🤔

tests/test_resource.py
```diff
@@ -330,6 +317,15 @@ def test_options_request(self):
             ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
         }
 
+    @patch.object(CORSResource, "resp_headers", return_value={})
+    def test_not_allow_oring_in_header(self, resp_headers_mock: MagicMock) -> None:
```
redlickigrzegorz, line 321: typo: _"ori**gin**"_

redlickigrzegorz: Please, describe it better. I think `CORSResource` allows having origin in headers but value can be unacceptable.

tests/test_resource.py
```diff
@@ -330,6 +317,15 @@ def test_options_request(self):
...
+        an_event = defaultdict(MagicMock())
+        an_event["method"] = "GET"
+        an_event["headers"] = {"origin": "notlocalhost"}
+        resp = CORSResource(an_event, ["GET", "POST"], origins="localhost")()
+        assert resp.status_code >= 400
```
redlickigrzegorz, line 326: This assertion gives a lot of options, please, choose one status code 😉

tests/test_resource.py
```diff
@@ -330,6 +317,15 @@ def test_options_request(self):
             ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
         }
 
+    @patch.object(CORSResource, "resp_headers", return_value={})
```
redlickigrzegorz, line 320: Why did you mock that? Isn't it better to check the result instead?

tests/test_resource.py
```diff
@@ -338,15 +334,16 @@ def setup_method(self, _test_method, _init_mock: MagicMock) -> None:
         self.resource = PaginatedCORSResource({}, [])
         self.resource.path = "/test/path"
         self.resource.urn = "/test/path"
-        req.query_params = MultiDict(
+        self.req = pytest.mark.usefixtures("sample_request")
```
redlickigrzegorz, line 337: This won't work as you expect. `pytest.mark.usefixtures` is a decorator, so you will get here the instance of one of `pytest` internal classes. Please. initialize the request normally 😉

tests/test_authz_collector.py
```diff
@@ -18,6 +18,9 @@ def test__init__(self) -> None:
         assert self.azc.resource_name == ""
         assert self.azc.guest_permissions == {}
 
+    def test__repr__(self) -> None:
+        assert str(self.azc) == "<AuthzCollector resource_name=>"
```
redlickigrzegorz, line 22: You should use `repr()` instead of `str()` - I think, I wrote a comment about that before 🤔

tests/test_authz_authorizer.py
```diff
@@ -1,35 +1,29 @@
...
+from lbz.authz.authorizer import Authorizer, ALL, ALLOW, DENY, LIMITED_ALLOW
 from lbz.exceptions import PermissionDenied, SecurityRiskWarning, Unauthorized
 from tests import SAMPLE_PRIVATE_KEY, EXPECTED_TOKEN
 
 
 # pylint: disable=too-many-public-methods
-class TestAuthorizer:
-    def setup_method(self):
+class TestAuthorizerSetupClass:
+    @pytest.fixture(autouse=True)
```
redlickigrzegorz, line 15: ditto: wrong usage of `pytest` fixtures

tests/test_authz_authorizer.py
```diff
@@ -1,35 +1,29 @@
...
-from lbz.authz import Authorizer, ALL, ALLOW, DENY, LIMITED_ALLOW
+from lbz.authz.authorizer import Authorizer, ALL, ALLOW, DENY, LIMITED_ALLOW
 from lbz.exceptions import PermissionDenied, SecurityRiskWarning, Unauthorized
 from tests import SAMPLE_PRIVATE_KEY, EXPECTED_TOKEN
 
 
 # pylint: disable=too-many-public-methods
-class TestAuthorizer:
-    def setup_method(self):
+class TestAuthorizerSetupClass:
```
redlickigrzegorz, line 14: You need to prepare better (more descriptive) names for the tests classes or just keep all tests in one as it was before 😉

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+    def test__repr__(self) -> None:
         assert repr(self.authz) == (
             "Authorizer(auth_jwt=<jwt>, "
             "resource_name='test_resource', "
             "permission_name='permission_name')"
         )
 
-    def test__set_policy_w_scope(self):
+    def test_validate_one_with_expired(self) -> None:
+        expiration_negative = int((datetime.utcnow() - timedelta(seconds=1)).timestamp())
```
redlickigrzegorz, line 43: What would you say for `expired_timestamp` Instead? 😉

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        with pytest.raises(Unauthorized):
+            self._make_authorizer(
+                {
+                    **self.token_payload,
+                    "allow": {"test_resource": {"permission_name": {"allow": ALL}}},
```
redlickigrzegorz, line 48: If you test here checking of expiration, why do you need to change permissions?

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        self.authz.deny = {"test_resource": {"permission_name": "self"}}
+        self.authz._check_deny()  # pylint: disable=protected-access
+        assert self.authz.denied_resource == "self"
+
+    def test__set_policy_w_scope(self) -> None:
         self.authz._set_policy(  # pylint: disable=protected-access
-            "", {"allow": "Lambda", "deny": "Lambda"}
+            "",
+            {
+                **self.token_payload,
```
redlickigrzegorz, line 155: Why do you need `self.token_payload` for `policy_override` - this test worked before 🤔

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+    def setup_method(self, base_auth_payload) -> None:
+        # pylint: disable=attribute-defined-outside-init
+        self.token_payload = base_auth_payload
+        self.authz = self._mocked_make_authorizer(self.token_payload)
+
+    def test_wrong_jwt_authz_payload_raises_permission_denied(self) -> None:
+        with pytest.raises(PermissionDenied):
+            self._mocked_make_authorizer({})
+
+    def test_check_deny_res(self) -> None:
```
redlickigrzegorz, line 71: _FYI:_ Changing the order of the tests when you are making multiple "fixes" in-between makes review much much harder.

tests/test_authz_authorizer.py
```diff
@@ -144,70 +250,16 @@ def test_missing_ref(self):
...
-        self.authz.denied_resource = {"city": "warszawa"}
-        assert self.authz.restrictions == {"allow": ALL, "deny": {"city": "warszawa"}}
-
-    def test_sign_authz(self):
+    def test_sign_authz(self) -> None:
         token = Authorizer.sign_authz({"allow": {ALL: ALL}, "deny": {}}, SAMPLE_PRIVATE_KEY)
         assert token == EXPECTED_TOKEN
+
+    def test_sign_authz_not_a_dict_error(self) -> None:
+        with pytest.raises(ValueError):
```
redlickigrzegorz, line 258: I think it would be good to check the message of this Exception 😉

tests/test_authz_authorizer.py
```diff
@@ -144,70 +250,16 @@ def test_missing_ref(self):
...
+            Authorizer.sign_authz({"allow": {ALL: ALL}, "deny": {}}, "")
```
redlickigrzegorz, line 259: Also, to make that more visible, I would keep only needed values:
```python
Authorizer.sign_authz({}, private_key_jwk="")
```

tests/test_authz_authorizer.py
```diff
@@ -144,70 +250,16 @@ def test_missing_ref(self):
...
+
+    def test_sign_authz_no_kid_error(self) -> None:
+        prive_key = deepcopy(SAMPLE_PRIVATE_KEY)
+        del prive_key["kid"]
+        with pytest.raises(ValueError):
+            Authorizer.sign_authz({"allow": {ALL: ALL}, "deny": {}}, prive_key)
```
redlickigrzegorz, line 265: The same here, you can simplify it and make more direct:
```python
key_without_kid = {}

with pytest.raises(ValueError):
    Authorizer.sign_authz({}, private_key_jwk=key_without_kid)
```

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        assert self.authz.outcome == ALLOW
+        assert self.authz.restrictions == {"allow": "*", "deny": None}
+
+    def test_check_allow_fail_all(self) -> None:
+        self.authz.allow = None
+        with pytest.raises(PermissionDenied):
+            self.authz._check_allow_and_set_resources()  # pylint: disable=protected-access
+            assert self.authz.outcome == ALLOW
+
+    def test_check_allow_domain(self) -> None:
```
redlickigrzegorz, line 119: It looks like this test is checking the access to the resource, not the domain.

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        self.authz.allow = {"test_resource": ALL}
+        self.authz._check_allow_and_set_resources()  # pylint: disable=protected-access
```
redlickigrzegorz, line 121: Why do you test a protected method? That can be used differently inside the class. I think testing `check_access()` makes the most sense.

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        assert self.authz.outcome == ALLOW
+
+    def test_check_deny_any(self) -> None:
+        self.authz.deny = {ALL: ALL}
+        with pytest.raises(PermissionDenied):
+            self.authz._check_deny()  # pylint: disable=protected-access
+        assert self.authz.outcome == DENY
+
+    def test_check_access_check_deny(self) -> None:
+        authz = self._mocked_make_authorizer({**self.token_payload, "deny": {ALL: ALL}})
```
redlickigrzegorz, line 88: If you mocked `decode_jwt`, the whole `self.token_payload` seems to be unneeded here 🤔 

Especially that you want to test the `deny` policy.

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        with pytest.raises(PermissionDenied):
+            authz.check_access()
+
+    def test_check_access_outcome_deny(self) -> None:
```
redlickigrzegorz, line 92: Please, use more words in the test names. Now, without checking the logic, it is hard to understand that correctly.

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+        authz = self._mocked_make_authorizer({**self.token_payload, "allow": {ALL: None}})
+        with pytest.raises(PermissionDenied):
+            authz.check_access()
+
+    def test__check_resource(self) -> None:
+        with patch.object(self.authz, "_deny_if_all", new_callable=MagicMock()) as mocked_deny:
+            self.authz._check_resource({"res": "xxx"})  # pylint: disable=protected-access
+            mocked_deny.assert_has_calls(
```
redlickigrzegorz, line 100: This whole test is wrongly designed:
1. `_deny_if_all` is a very very simple method, it doesn't have to be mocked.
2. `_check_resource` is a protected method - shouldn't be tested.
3. By checking that one protected method triggered another protected method three times, you have frozen the implementation, not the result of the code.

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+
+
+# pylint: disable=too-many-public-methods
+class TestAuthorizerSetupMethod:
+    @staticmethod
+    def _mocked_make_authorizer(token_payload: dict) -> Authorizer:
+        with patch("lbz.authz.authorizer.decode_jwt", lambda _: token_payload):
+            return Authorizer("xx", "test_resource", "permission_name")
+
+    @pytest.fixture(autouse=True)
```
redlickigrzegorz, line 61: ditto: wrong usage of `pytest` fixtures

tests/test_dev_server.py
```diff
@@ -2,38 +2,37 @@
...
-    @add_route("/", method="GET")
-    def list(self):
-        return Response({"message": "HelloWorld"})
 
-    @add_route("/t/{id}", method="GET")
-    def get(self):
-        return Response({"message": "HelloWorld"})
+def test_my_dev_server_no_cls_raises() -> None:
+    with pytest.raises(TypeError):
+        MyLambdaDevHandler.cls()  # pylint: disable=no-value-for-parameter
```
redlickigrzegorz, line 17: That is a totally wrong usage of `cls` property 🤔  I don't see value from checking that this way. Can you explain it?

tests/test_dev_server.py
```diff
@@ -2,38 +2,37 @@
...
 
 
 class MyLambdaDevHandlerHelloWorld(MyLambdaDevHandler):
-    cls = HelloWorld
+    cls = SimpleTestResource
```
redlickigrzegorz, line 21: I would keep that with the `HelloWorld` as it was implemented before. The more complicated code written only for tests makes tests more complicated - keep it simple 😉  Most of the things that are implemented inside this resource should be unit tested anyway.

tests/test_dev_server.py
```diff
@@ -49,10 +48,87 @@ def test_my_lambda_dev_handler():
     assert path == "/t/{id}"
     assert params == {"id": "123"}
 
+    path, params = handler._get_route_params("/?=1&xx=2")  # pylint: disable=protected-access
+    assert path == "/"
+    assert params is None
+
+    with pytest.raises(ValueError):
+        handler._get_route_params("")  # pylint: disable=protected-access
```
redlickigrzegorz, line 56: Keep the test cases separately.

tests/test_dev_server.py
```diff
@@ -49,10 +48,87 @@ def test_my_lambda_dev_handler():
...
+
+
+@pytest.mark.parametrize(
+    "method", ["do_GET", "do_PATCH", "do_POST", "do_PUT", "do_DELETE", "do_OPTIONS"]
+)
+def test_handle_request(method) -> None:
+    with patch("socket.socket") as msocket:
+        msocket.makefile = lambda a, b: io.BytesIO(b"GET / HTTP/1.1\r\n")
+        msocket.rfile.close = lambda: 0
```
redlickigrzegorz, line 65: The `msocket` variable should become a separate fixture. Mostly, to describe that better but also, to reduce the duplication. The biggest value of the tests is the documentation which can be checked anytime by anyone - so, we need to try to make the tests simple 😉

tests/test_dev_server.py
```diff
@@ -49,10 +48,87 @@ def test_my_lambda_dev_handler():
...
     assert dev_serv.server_address == ("localhost", 8000)
     assert dev_serv.port == 8000
     assert dev_serv.address == "localhost"
     assert issubclass(dev_serv.my_handler, MyLambdaDevHandler)
+
+
+def test_my_dev_server_run() -> None:
+    dev_serv = MyDevServer(MyLambdaDevHandlerHelloWorld)
+    dev_serv.start()
+    assert dev_serv.is_alive()
```
redlickigrzegorz, line 132: Don't you test that in a separate test module? `test_dev_server_running`? 🤔  I think I am getting a little lost. You are mixing the running server with the one which is partially mocked, and with that poor naming of the tests and duplicate part of the code, this won't be maintainable.

I wrote it before, but part of the tests for this library is more like integration tests which should be kept separately and better implemented/documented.

tests/test_dev_server.py
```diff
@@ -49,10 +48,87 @@ def test_my_lambda_dev_handler():
...
+    byte_string = (
+        b'POST /p HTTP/1.1\r\nContent-Length: 8\r\nContent-Type: application/json\r\n\r\n{"a": 1}'
+    )
+    with patch.object(MyLambdaDevHandler, "_send_json") as mocked_send:
+        with patch("socket.socket") as msocket:
+            msocket.makefile = lambda a, b: io.BytesIO(byte_string)
+            msocket.rfile.close = lambda: 0
+            MyLambdaDevHandlerHelloWorld(msocket, ("127.0.0.1", 8888), BaseServer)
+
+        mocked_send.assert_has_calls(
```
redlickigrzegorz, line 109: In both `test_handle_txt` and `test_handle_json`, you are checking if one of the mocked/protected methods was triggered according to the needs. In general, that is not what the test should test 😐  You should rather compare the results which you are getting.

tests/test_dev_misc.py
```diff
@@ -1,12 +1,68 @@
...
 from hashlib import sha1
 
-from lbz.dev.misc import EVENT_TEMPLATE
+import pytest
 
+from lbz.dev.misc import EVENT_TEMPLATE, Event
 
-def hash_string(string):
+
+def hash_string(string) -> None:
```
redlickigrzegorz, line 9: This function does not return `None`.

redlickigrzegorz (commented):
Ok, I think that is the last review that I have done for this enormous pull request. My recommendation would be to not touch this branch at all. A much simpler and quicker solution will be to split those changes into several **SMALLER** pull requests and merge them separately 😉

tests/conftest.py
```diff
@@ -1,32 +1,110 @@
...
 
+from lbz.authentication import User
+from lbz.authz.authorizer import Authorizer
 from lbz.dev.misc import Event
 from lbz.request import Request
+from tests import SAMPLE_PRIVATE_KEY
+from tests.fixtures.rsa_pair import SAMPLE_PUBLIC_KEY
+from tests.utils import encode_token, allowed_audiences
 
+req = Request(
```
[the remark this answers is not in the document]
pdyba: Request in consider unmutable (except methods but they are not changing the data)

tests/conftest.py
```diff
@@ -1,32 +1,110 @@
...
+    return USERNAME
+
+
+@pytest.fixture(autouse=True)
+def user_cogniot() -> dict:
+    return COGNITO_USER
+
+
+@pytest.fixture(autouse=True)
+def user() -> User:
```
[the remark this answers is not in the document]
pdyba: Actually in orded to makes test faster we need to fo all jwt related, but we might store them in more "private way"

tests/test_authz_authorizer.py
pdyba: becouse here we are testing the authorize module ergo all its private and magick methods.

tests/test_authz_authorizer.py
pdyba: as above it this module it need to be checked.

tests/test_authz_authorizer.py
pdyba: actually its correct - in this place it enforcing using fixutres as this is not test_ method.

tests/conftest.py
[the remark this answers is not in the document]
redlickigrzegorz: Unfortunately, there are places where this rule is broken e.g. [test_resource.py](https://github.com/pdyba/lambdalizator/blob/master/tests/test_resource.py#L341). By using fixtures, we have a much clearer situation because it is impossible to change that 😉

tests/conftest.py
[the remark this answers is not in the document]
redlickigrzegorz: As we agreed offline, the `pytest` fixtures with the scope set to _session_ might be the solution in this situation.

tests/test_authz_authorizer.py
redlickigrzegorz: But I am speaking about the general approach to testing - if something has been marked as protected, you shouldn't use it outside the class/module. If you need to do this, that starts smelling like a problem either in the test approach or the code in general 😬 

Of course, you can simply mark that as public (not recommended) or test something that is public and what will be really used by the outside world 😉

tests/test_authz_authorizer.py
redlickigrzegorz: It depends on what you want to achieve, right now, with these tests, there is no chance to change the code without getting errors from tests. In general, that is ok but it blocks the code too much - it won't give the possibility to easily refactor the code.

By tests, we should check everything that is important meaning e.g. the result, but it does not matter how the result has been achieved 😉

tests/test_authz_authorizer.py
redlickigrzegorz: Yes, it works as you want, but that is only a side effect. In reality, you have made a fixture based on this method - that is something against the pattern of how the `pytest` fixtures should be used. Please, use fixtures as they should be used, so everywhere directly.

tests/test_authz_authorizer.py
pdyba: agreed to disagree offline.

tests/test_authz_authorizer.py
pdyba: resource == domain 
rented

tests/test_authz_authorizer.py
pdyba: agreed to disagree offline.

tests/test_authz_authorizer.py
redlickigrzegorz: Confusing, I must say 😉

tests/test_authz_authorizer.py
redlickigrzegorz: As below: we will come back to this topic in the future, because it is an important thing when designing the code 😉

PR #42, opened 2021-09-07
Authz fix

pdyba:


lbz/authz/decorators.py
```diff
@@ -18,7 +18,9 @@ def check_permission(resource: Resource, permission_name: str) -> dict:
     Raises if not.
     """
     authorization_header = resource.request.headers.get("Authorization", "")
-    guest_authorization_policy = resource.get_guest_authorization()
+    guest_authorization_policy = (
+        resource.get_guest_authorization() if not authorization_header else None
```
redlickigrzegorz, line 22: I liked [the version before introducing this bug](https://github.com/pdyba/lambdalizator/blob/0ffb9f4187d7b6c5fb1935ab17858df49ca0de11/lbz/authz.py#L177) more. It is easier to read when we don't check the same thing multiple times (a little bit updated below):
```python
guest_authorization_policy = None
if not (authorization_header := resource.request.headers.get("Authorization", "")):
    if not (guest_authorization_policy := resource.get_guest_authorization()):
        raise Unauthorized("Authorization header missing or empty")
```

redlickigrzegorz: What about the tests for this change? All of them are passing as before which doesn't mean it is good because we can introduce a similar bug in the future 😐

tests/test_authz_decorator.py
```diff
@@ -100,3 +104,47 @@ def handler(self, restrictions):
         auth_header = Authorizer.sign_authz({"allow": {"x": "*"}, "deny": {}}, SAMPLE_PRIVATE_KEY)
         resp = XResource({**event, "headers": {"authorization": auth_header}})()
         assert resp.status_code == HTTPStatus.FORBIDDEN
+
+    def test_check_permission(self) -> None:
```
redlickigrzegorz, line 108: You are in the `TestAuthorizationDecorator` class, you shouldn't put here tests for the `check_permission` function. Especially that they look like duplicates of the tests which I see above 🤔

redlickigrzegorz: Also, I don't see the test which proves that the bug is fixed. I think the simplest solution here is to follow the currently chosen pattern and test the fix via the `authorization` decorator, and improve in the future.

redlickigrzegorz: Something like:
```python
@patch.dict(environ, env_mock)
def test_authorization_header_takes_priority_over_guest_policy_when_both_present(self, *_args):
    class XResource(Resource):
        @add_route("/")
        @authorization()
        def handler(self, restrictions):
            return Response("x")

        @staticmethod
        def get_guest_authorization() -> dict:
            return {"allow": {}, "deny": {}}

    auth_header = Authorizer.sign_authz({"allow": "*", "deny": {}}, SAMPLE_PRIVATE_KEY)
    resp = XResource({**event, "headers": {"authorization": auth_header}})()
    assert resp.status_code == 200
```

Runt this test with both the old and the new code, you will see that it is passing only after the fix 😉

redlickigrzegorz (approved):
I have approved this PR in advance but please, replace these new tests with one which proves the correctness of the fix (I proposed one) 😉

PR #43, opened 2021-09-10
0.3.16 stable

pdyba:


tests/conftest.py
```diff
@@ -1,32 +1,99 @@
...
+    }
+
+
+@pytest.fixture(scope="session")
+def base_auth_payload() -> dict:
+    return _base_auth_payload()
+
+
+@pytest.fixture(scope="session")
+def auth_header() -> str:
```
redlickigrzegorz, line 62: Do you use this fixture anywhere? Actually, I don't see any usage 🤔

tests/conftest.py
```diff
@@ -1,32 +1,99 @@
...
         resource_path="/",
         method="GET",
         headers={},
         path_params={},
         query_params={},
+        body=sample_request.to_dict(),
+    )
+
+
+def _base_auth_payload() -> dict:
```
redlickigrzegorz, line 46: Shouldn't it be a `full_access_authz_payload`? You are granting access to everything here, so it shouldn't be simple base, I think 😉

tests/conftest.py
```diff
@@ -1,32 +1,99 @@
...
+@pytest.fixture(scope="session")
+def auth_header() -> str:
+    return Authorizer.sign_authz(
+        {**_base_auth_payload(), "allow": {"test_res": {"perm-name": {"allow": "*"}}}, "deny": {}},
+        SAMPLE_PRIVATE_KEY,
     )
+
+
+@pytest.fixture(scope="session")
+def user_username() -> str:
```
redlickigrzegorz, line 70: It appears that all of the following fixtures are also not used 🤷‍♀️

tests/test_authentication.py
```diff
@@ -23,14 +23,14 @@ def setup_class(self):
         ):
             self.cognito_user = {
                 "cognito:username": str(uuid4()),
-                "custom:id": str(uuid4()),
                 "email": f"{str(uuid4())}@{str(uuid4())}.com",
+                "custom:id": str(uuid4()),
                 "custom:1": str(uuid4()),
                 "custom:2": str(uuid4()),
+                "aud": allowed_audiences[0],  # moved due pylint and minimising scope of PR
```
redlickigrzegorz, line 30: This comment makes sense only in the pull request where I see exactly what was introduced, but in the code, you should rather avoid things like this or point directly to [#214](https://github.com/PyCQA/pylint/issues/214).

tests/test_authz_authorizer.py
```diff
@@ -1,35 +1,29 @@
...
+from lbz.authz.authorizer import Authorizer, ALL, ALLOW, DENY, LIMITED_ALLOW
 from lbz.exceptions import PermissionDenied, SecurityRiskWarning, Unauthorized
 from tests import SAMPLE_PRIVATE_KEY, EXPECTED_TOKEN
 
 
 # pylint: disable=too-many-public-methods
-class TestAuthorizer:
-    def setup_method(self):
+class TestAuthorizerSetupClass:
+    @pytest.fixture(autouse=True)
```
redlickigrzegorz, line 15: As I mentioned multiple times in #40: that is wrong usage of `pytest` fixtures, please get rid of that 😉

tests/test_authz_authorizer.py
```diff
@@ -38,37 +32,161 @@ def test__init__(self):
...
+    def test__repr__(self) -> None:
         assert repr(self.authz) == (
             "Authorizer(auth_jwt=<jwt>, "
             "resource_name='test_resource', "
             "permission_name='permission_name')"
         )
 
-    def test__set_policy_w_scope(self):
+    def test_validate_one_with_expired(self) -> None:
+        expiration_negative = int((datetime.utcnow() - timedelta(seconds=1)).timestamp())
```
redlickigrzegorz, line 43: I also see that some of the other comments from #40 were not applied - I don't see any point in repeating them here, but I'd like to see the fixes 😉

lbz/request.py
```diff
@@ -68,12 +68,14 @@ def json_body(self) -> Optional[dict]:
         if content_type is None:
             return None
         if content_type.startswith("application/json"):
+            if isinstance(self._body, dict):
```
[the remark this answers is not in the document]
redlickigrzegorz: Also, this is the most important change to make the whole lib stable again but all the tests are passing with the both old and new versions. So, in general, there is no proof that it works and does not break anything else 😬 

With all the things I pointed around, I think we cannot follow the cowboy coding 😕

lbz/request.py
```diff
@@ -52,11 +52,11 @@ def _decode_base64(encoded: Union[str, bytes]) -> bytes:
         return base64.b64decode(encoded)
 
     @property
-    def raw_body(self) -> Union[bytes, str]:
+    def raw_body(self) -> Union[bytes, str, dict]:
         if not self._raw_body and self._body is not None:
-            if self._is_base64_encoded:
+            if self._is_base64_encoded and isinstance(self._body, bytes):
```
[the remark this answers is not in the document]
pdyba: actually this is correct now, it was also correct before if payload would be an str it would not trigger as the if was on _is_bas64 flag which should only be in place with bytes.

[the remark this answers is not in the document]
redlickigrzegorz: But the `self._body` is coming directly from the event so, in the case of e.g. API Gateway, it will be always string. I would say, we cannot rely here only on `_is_base64_encoded` flag.

Anyway, you have improved that, so it looks like the problem is solved 😉

tests/conftest.py
```diff
@@ -1,32 +1,130 @@
...
+    full_access_authz_payload,  # pylint: disable=redefined-outer-name
+) -> str:
+    return Authorizer.sign_authz(
+        full_access_authz_payload,
+        SAMPLE_PRIVATE_KEY,
+    )
+
+
+@pytest.fixture(scope="session")
+def limited_access_auth_header(
```
redlickigrzegorz, line 73: As I see, this fixture is not used at all _#YAGNI_

tests/conftest.py
redlickigrzegorz: It started to be used, so my previous comment is no longer valid, I guess.

tests/test_authz_authorizer.py
```diff
@@ -1,130 +1,241 @@
...
+        with pytest.raises(PermissionDenied):
+            authz._check_deny()  # pylint: disable=protected-access
+        assert authz.outcome == DENY
+
+    def test_check_access_check_deny_outcome(self) -> None:
+        authz = self._make_mocked_authorizer({"allow": {}, "deny": {ALL: ALL}})
+        with pytest.raises(PermissionDenied):
+            authz.check_access()
+
+    def test_check_access_outcome_deny(self, full_access_authz_payload) -> None:
```
redlickigrzegorz, line 85: You have two new tests here:
- test_check_access_check_deny_outcome
- test_check_access_outcome_deny

Unfortunately, they are named almost the same, which doesn't help to understand what you wanted to achieve and won't help to develop that in the future 😉  As I mentioned before, please, you longer, self-descriptive names for the tests.

tests/test_authz_authorizer.py
```diff
@@ -1,130 +1,241 @@
...
+    def test_check_allow_fail_all(self, full_access_authz_payload) -> None:
+        authz = self._make_mocked_authorizer(full_access_authz_payload)
+        authz.allow = None
+        with pytest.raises(PermissionDenied):
+            authz._check_allow_and_set_resources()  # pylint: disable=protected-access
+            assert authz.outcome == ALLOW
+
+    def test_check_allow_resource(self, full_access_authz_payload) -> None:
+        authz = self._make_mocked_authorizer(full_access_authz_payload)
+        authz.allow = {"test_resource": ALL}
```
redlickigrzegorz, line 117: Can't you just create an object you need for the test instead of changing it one line below? 🤔

tests/test_authz_authorizer.py
```diff
@@ -133,81 +244,27 @@ def test_refs(self):
...
+
+    def test_sign_authz_not_a_dict_error(self) -> None:
+        with pytest.raises(ValueError) as execinfo:
+            Authorizer.sign_authz({}, private_key_jwk="")
+            assert str(execinfo.value) == "private_key_jwk must be a jwk dict"
+
+    def test_sign_authz_no_kid_error(self) -> None:
+        with pytest.raises(ValueError) as execinfo:
+            Authorizer.sign_authz({}, private_key_jwk={})
+            assert str(execinfo.value) == "private_key_jwk must have the 'kid' field"
```
redlickigrzegorz, line 270: Pytest provides the possibility to check the error message in a more elegant way:
```python
with pytest.raises(ValueError, match="private_key_jwk must have the 'kid' field"):
    Authorizer.sign_authz({}, private_key_jwk={})
```

redlickigrzegorz (approved):
During the last Code Review session, I have put a couple of additional comments which should be considered to improve this code in general (especially the tests which quality is still a little questionable).

I approved that but I am pretty sure we will return to this module to refactor it 😉

tests/test_request.py
```diff
@@ -78,62 +47,74 @@ def test__decode_base64_str(self):
...
-    def test_json_body_dict(self):
-        self.resp._body = '{"x": "abcx"}'  # pylint: disable=protected-access
-        assert self.resp.json_body == {"x": "abcx"}
 
-    def test_json_body_json(self):
-        self.resp._json_body = None  # pylint: disable=protected-access
-        self.resp._body = '{"x": "abcx"}'  # pylint: disable=protected-access
-        assert self.resp.json_body == {"x": "abcx"}
-        assert self.resp._json_body == {"x": "abcx"}  # pylint: disable=protected-access
+class TestRequestJsonBody:
```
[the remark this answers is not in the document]
pdyba: Classes groups test here for specific public method - for me it was easier to read even if there is no performence/code benefit at all.

tests/test_request.py
```diff
@@ -78,62 +47,74 @@ def test__decode_base64_str(self):
...
+    def test_raw_body_base64(self, sample_request):
+        sample_request._body = b"asdasdsd"  # pylint: disable=protected-access
+        sample_request._is_base64_encoded = True  # pylint: disable=protected-access
+        assert sample_request.raw_body == b"j\xc7Z\xb1\xdb\x1d"
 
-    def test_raw_body_bytes(self):
-        self.resp._body = b"asdasdsd"  # pylint: disable=protected-access
-        assert self.resp.raw_body == b"asdasdsd"
+    def test_raw_body_bytes(self, sample_request):
+        sample_request._body = b"asdasdsd"  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
pdyba: add todo

lbz/request.py
```diff
@@ -52,32 +52,42 @@ def _decode_base64(encoded: Union[str, bytes]) -> bytes:
         return base64.b64decode(encoded)
 
     @property
-    def raw_body(self) -> Union[bytes, str]:
+    def raw_body(self) -> Optional[Union[bytes, dict]]:
         if not self._raw_body and self._body is not None:
```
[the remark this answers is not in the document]
pdyba: actually it should be:
if self._raw_body is None...
is **not** None would be the oposit of what we want ;)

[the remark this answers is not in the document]
redlickigrzegorz: Yes, you are absolutely right 🙂

PR #44, opened 2021-09-17
0.3.16a

pdyba:


version
```diff
@@ -1 +1 @@
-0.3.16
\ No newline at end of file
+0.3.16a
```
redlickigrzegorz, line 1: As we discussed offline, we should rethink the version number. In general, all the problems with `lbz` started after releasing #37 (_v0.3.14_). I know you removed some versions so maybe also this one should be treated the same way.

So, to not put here additional letters or other signs, we should use the first available version: _v0.3.19_ or think about increasing the version to `v0.4.0` 😉

redlickigrzegorz: No matter what you decide, that should be described in `CHANGELOG.md` 😉

redlickigrzegorz (approved):

PR #46, opened 2021-10-09
Jwt cases

pdyba:


tests/test_jwt_utils.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @patch.object(jwt, "get_unverified_header", return_value={})
+    def test_get_matching_missing_key(self, _get_unverified_header_mock: MagicMock) -> None:
+        with pytest.raises(Unauthorized):
+            get_matching_jwk("x")
+
+
+class TestDecodeJWT:
+    @patch("lbz.jwt_utils.get_matching_jwk", return_value={})
+    def test_missing_public_keys(self, get_matching_jwk_mock: MagicMock) -> None:
```
redlickigrzegorz, line 36: Based on the code, if public keys are missing, we should get the `RuntimeError` 🤔  Can you verify that?

tests/test_jwt_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        decoded_jwt_data = decode_jwt(full_access_auth_header)
+        assert decoded_jwt_data == full_access_authz_payload
+
+    def test_expired_jwt(self) -> None:
+        iat = int((datetime.utcnow() - timedelta(hours=12)).timestamp())
+        exp = int((datetime.utcnow() - timedelta(hours=6)).timestamp())
+        token_payload = {
+            "exp": exp,
+            "iat": iat,
+            "iss": "test-issuer",
```
redlickigrzegorz, line 58: What about the `aud`? I think it would be good to provide all values even in tests 😉

tests/test_jwt_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        get_matching_jwk_mock.assert_called_once_with("x")
+
+    @patch("lbz.jwt_utils.get_matching_jwk", return_value={})
+    def test_invalid_type(self, get_matching_jwk_mock: MagicMock) -> None:
+        msg = "error occurred during decoding"
+        with pytest.raises(RuntimeError, match=msg):
+            decode_jwt({"a"})
+        get_matching_jwk_mock.assert_called_once_with({"a"})
+
+    def test_proper_jwt(self, full_access_authz_payload, full_access_auth_header) -> None:
```
redlickigrzegorz, line 48: I know `mypy` is not configured for tests but it will be easier to enable that if you don't skip type annotation in the new code 😉

tests/test_jwt_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        iat = int(datetime.utcnow().timestamp())
+        exp = int((datetime.utcnow() + timedelta(hours=6)).timestamp())
+        token_payload = {"exp": exp, "iat": iat, "iss": "test-issuer", "aud": "test"}
+        jwt_token = Authorizer.sign_authz(token_payload, SAMPLE_PRIVATE_KEY)
+        with pytest.raises(Unauthorized):
+            decode_jwt(jwt_token)
+
+
+@patch("lbz.jwt_utils.PUBLIC_KEYS", [])
+class TestDecodeJWTMissingKEy:
```
redlickigrzegorz, line 89: I don't see any point in separating those three tests from the main class 🤔  Can you put them into `TestDecodeJWT` or explain?

tests/test_jwt_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        self, get_matching_jwk_mock: MagicMock, decode_mock: MagicMock
+    ) -> None:
+        with pytest.raises(Unauthorized):
+            decode_jwt("x")
+        get_matching_jwk_mock.assert_called_once_with("x")
+        decode_mock.assert_not_called()
+
+
+class TestDecodeJWTAudiencesMissMatch:
+    def test_missing_audiences(self) -> None:
```
redlickigrzegorz, line 79: I think you should improve all names of the tests in this module. You have two test methods named this way but they are testing a little bit different things.

tests/test_jwt_utils.py
```diff
@@ -0,0 +1,93 @@
...
+            "aud": "test-audience",
+        }
+        jwt_token = Authorizer.sign_authz(token_payload, SAMPLE_PRIVATE_KEY)
+        with pytest.raises(Unauthorized, match="Your token has expired. Please refresh it."):
+            decode_jwt(jwt_token)
+
+    @patch.object(jwt, "decode")
+    @patch("lbz.jwt_utils.get_matching_jwk", return_value={})
+    @patch("lbz.jwt_utils.ALLOWED_AUDIENCES", [])
+    def test_empty_allowed_audiences(
```
redlickigrzegorz, line 69: Shouldn't such a situation be caught earlier? I mean to treat that the same way as empty `PUBLIC_KEYS` 🤔

redlickigrzegorz (approved):

tests/test_jwt_utils.py
pdyba: it should - fixed that but Testing that is hard.

lbz/jwt_utils.py
```diff
@@ -25,6 +25,9 @@
 if any("kid" not in public_key for public_key in PUBLIC_KEYS):
     raise ValueError("One of the provided public keys doesn't have the 'kid' field")
 
+if not ALLOWED_AUDIENCES:
+    raise ValueError("ALLOWED_AUDIENCES where not defined")
```
redlickigrzegorz, line 29: Typo: "_... were ..._"

lbz/jwt_utils.py
```diff
@@ -25,6 +25,9 @@
 if any("kid" not in public_key for public_key in PUBLIC_KEYS):
     raise ValueError("One of the provided public keys doesn't have the 'kid' field")
 
+if not ALLOWED_AUDIENCES:
```
redlickigrzegorz, line 28: Above, you check only the correctness of `PUBLIC_KEYS` if present but this new condition makes `lbz` unusable without setting this variable 🤔  My suggestion would be to return to that to make _v0.4.x_ work as before 😉

PR #47, opened 2021-10-10
Ahutz inharitance

pdyba:


tests/conftest.py
```diff
@@ -126,3 +138,51 @@ def sample_request_with_user(user) -> Request:  # pylint: disable=redefined-oute
...
+        method="GET",
+        headers={"authorization": limited_access_auth_header},
+        path_params={},
+        query_params={},
+        body={},
+    )
+
+
+@pytest.fixture()
+def sample_resoruce_with_authorization() -> Resource:
```
redlickigrzegorz, line 173: This fixture returns `Type[Resource]`.

lbz/dev/server.py
```diff
@@ -13,6 +14,9 @@
 from lbz.dev.misc import Event
 from lbz.resource import Resource
 
+if not environ.get("LBZ_DEBUG_MODE"):
+    environ["LBZ_DEBUG_MODE"] = "True"
```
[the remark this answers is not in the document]
pdyba: fixed but does not actually matter

tests/test_authz_authorizer.py
```diff
@@ -74,8 +74,14 @@ def test_check_deny_any(self, full_access_authz_payload) -> None:
             authz._check_deny()  # pylint: disable=protected-access
         assert authz.outcome == DENY
 
-    def test__check_access__outcome_deny_bcs_deny_all(self) -> None:
-        authz = self._make_mocked_authorizer({"allow": {}, "deny": {ALL: ALL}})
+    def test__check_access__outcome_deny_bcs_deny_all(self, full_access_authz_payload) -> None:
+        authz = self._make_mocked_authorizer(
+            {
+                **full_access_authz_payload,
+                "allow": {},
```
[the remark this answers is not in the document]
pdyba: added comment but that was done to get iss/aud etc.

tests/test_authz_authorizer.py
```diff
@@ -214,13 +221,19 @@ def test_validate_one(self, full_access_authz_payload) -> None:
...
-        authorizer.check_access()
-        assert authorizer.outcome == ALLOW
+
+    def test_validate_one_missing_exp_exceptio(self) -> None:
+        with pytest.raises(SecurityError, match="'iss'"):
+            self._make_mocked_authorizer(
+                {
+                    "allow": "*",
+                    "deny": {},
+                    "exp": int((datetime.utcnow() + timedelta(hours=6)).timestamp()),
```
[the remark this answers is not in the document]
pdyba: Leaving it as is provides a little bit more info then magic int. ;)

tests/conftest.py
```diff
@@ -126,3 +138,51 @@ def sample_request_with_user(user) -> Request:  # pylint: disable=redefined-oute
         is_base64_encoded=False,
         uri_params={},
     )
+
+
+@pytest.fixture()
+def sample_event_with_full_access_auth_header(
```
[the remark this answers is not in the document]
redlickigrzegorz: I know `sample_event_with_limited_access_auth_header` is used in more places but still, replacing it with the similar line I mentioned above, will simplify the tests - you won't have to check the fixture and understand magic parameterization.

Maybe it can be named _duplication_ but it is only _apparent duplication_ - there is no possibility to make everything generic. In general, in the tests, we should follow simplicity instead of trying to write not duplicate code (especially when we talk about one line 😉 ).

tests/test_authz_utils.py
```diff
@@ -2,51 +2,43 @@
...
+from lbz.exceptions import PermissionDenied, Unauthorized
+from lbz.authz.utils import has_permission, check_permission
 
 
-class SampleResource(Resource):
-    _name = "sample_resource"
-
-
-class TestAuthorizationUtils:
+class TestAuthorizationDecorator:
```
redlickigrzegorz, line 9: The name of the test class is not quite correct.

tests/test_authz_utils.py
```diff
@@ -2,51 +2,43 @@
...
-        sample_event["headers"]["authorization"] = authorization
+    def test_check_permission(
+        self, sample_event_with_limited_access_auth_header, sample_resoruce_with_authorization
+    ) -> None:
+        res_instance = sample_resoruce_with_authorization(
+            sample_event_with_limited_access_auth_header
+        )
+        assert check_permission(res_instance, "perm-name") == {"allow": "*", "deny": None}
+        with pytest.raises(PermissionDenied):
+            check_permission(res_instance, "garbage")
```
redlickigrzegorz, line 23: Those are two separate scenarios, you shouldn't put them together.

tests/test_authz_utils.py
```diff
@@ -2,51 +2,43 @@
...
+        indirect=True,
+    )
+    def test_has_permission(
+        self, sample_event_with_limited_access_auth_header, sample_resoruce_with_authorization
+    ) -> None:
+        res_instance = sample_resoruce_with_authorization(
+            sample_event_with_limited_access_auth_header
+        )
+        assert has_permission(res_instance, "perm-name")
+        assert not has_permission(res_instance, "garbage")
```
redlickigrzegorz, line 44: The same here, two separate scenarios, please, parametrize it or create separate tests for that.

redlickigrzegorz (approved):
I am approving this PR but still, that requires some additional work (based on the new comments but also the older ones). Most of the things which should be done relate to tests but without good tests we cannot talk about the quality so they are important anyway 😉

lbz/jwt_utils.py
```diff
@@ -9,25 +9,20 @@
...
 PUBLIC_KEYS: List[dict] = []
 ALLOWED_AUDIENCES = []
+ALLOWED_ISS = os.environ.get("ALLOWED_ISS", "")
 
 if allowed_pubkeys_str := os.environ.get("ALLOWED_PUBLIC_KEYS"):
     PUBLIC_KEYS.extend(json.loads(allowed_pubkeys_str)["keys"])
 if allowed_audiences_str := os.environ.get("ALLOWED_AUDIENCES"):
     ALLOWED_AUDIENCES.extend(allowed_audiences_str.split(","))
 
-if any("kid" not in public_key for public_key in PUBLIC_KEYS):
```
redlickigrzegorz, line 25: I would keep this condition (only this one) here. It confirms only if all public keys are correct. So, it can be done once at the very beginning because if `PUBLIC_KEYS` is empty, nothing bad will happen 😉

lbz/jwt_utils.py
```diff
@@ -59,10 +64,17 @@ def decode_jwt(auth_jwt_token: str) -> dict:
         msg = "Invalid configuration - no keys in the ALLOWED_PUBLIC_KEYS env variable"
         raise RuntimeError(msg)
 
+    if any("kid" not in public_key for public_key in PUBLIC_KEYS):
+        raise RuntimeError("One of the provided public keys doesn't have the 'kid' field")
+
+    if not ALLOWED_AUDIENCES:
+        raise RuntimeError("ALLOWED_AUDIENCES where not defined")
```
redlickigrzegorz, line 71: Typo: "_... were ..._"

tests/test_authz_authorizer.py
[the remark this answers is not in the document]
redlickigrzegorz: That's the thought standing behind my previous comment, we shouldn't create tests that can pass for different reasons. The environment should be prepared the way which represents the edge case which we want to test 😉 

Here, no matter if you have `"deny": {ALL: ALL}` or just `"deny": {}` this test will pass, so we can modify the code and this check won't point potential problems.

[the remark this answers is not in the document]
redlickigrzegorz: The same problem I see in the following tests 😬  So, we cannot be 100% sure that it works correctly.

tests/test_authz_decorator.py
```diff
@@ -1,167 +1,81 @@
...
-            @staticmethod
-            def get_guest_authorization() -> dict:
-                return {"allow": {}, "deny": {}}
-
-        auth_header = Authorizer.sign_authz({"allow": "*", "deny": {}}, SAMPLE_PRIVATE_KEY)
-        resp = XResource({**event, "headers": {"authorization": auth_header}})()
-        assert resp.status_code == 200
+    def test_inharitance_success(self, full_access_auth_header) -> None:
+        res_instance = GuestResource(
+            Event("/garbage2", "GET", headers={"authorization": full_access_auth_header})
```
redlickigrzegorz, line 79: This test does not confirm inheritance 😕  You should pass the Authorization header which does not have access to the `handler` method and thanks to inheritance it will get it anyway because it is declared in the guest authorization:
```python
def test_inheritance_success(self, limited_access_auth_header: str) -> None:
    # the limited access auth header grants access only to the `test_res`
    event = Event("/", "GET", headers={"authorization": limited_access_auth_header})
    res_instance = GuestResource(event)

    # no matter the user accesses, they will get the OK response because of guest permissions
    assert res_instance().status_code == HTTPStatus.OK
```

redlickigrzegorz (commented):
As I promised, I have checked the changes in tests done after my approval. In my opinion, they should be improved in the upcoming versions of the library (I left a couple of comments) 😉

PR #48, opened 2021-11-16
Additional cors headers

pdyba:


lbz/resource.py
```diff
@@ -135,14 +134,22 @@ class CORSResource(Resource):
         "X-Amz-Security-Token",
     ]
 
-    def __init__(self, event: dict, methods: List[str], origins: List[str] = None):
+    def __init__(
+        self,
+        event: dict,
+        methods: List[str],
+        origins: List[str] = None,
+        cors_headers: List[str] = None,
```
redlickigrzegorz, line 142: Still, I see some inconsistency here: we have already `methods` and `origins` in this class which in the end means allowed values. So, following this pattern, I think I would name this simply `headers` (meaning: allowed headers) 😉

redlickigrzegorz: Or, if we want to provide a new pattern, TODO would be welcome to adjust the rest of the arguments in the near future too.

tests/test_resource.py
```diff
@@ -330,6 +334,31 @@ def test_options_request(self):
...
+            "Access-Control-Allow-Headers": (
+                "Content-Type, X-Amz-Date, Authentication, Authorization, X-Api-Key, "
+                "X-Amz-Security-Token, X-PRN-KEY"
+            ),
+            "Access-Control-Allow-Methods": "GET, POST, OPTIONS",
+            ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
+        }
+
+    def test_cors_headers_env_request(self) -> None:
+        environ["CORS_HEADERS"] = "X-PRN-KEY,X-PRN-TOKEN"
```
redlickigrzegorz, line 350: The env variable should be unset after the finished job, otherwise, it can affect the next tests 😉 
```python
@patch.dict(environ, {"CORS_HEADERS": "X-PRN-KEY,X-PRN-TOKEN"})
def test_cors_headers_env_request(self) -> None:
    ...
```

tests/test_resource.py
```diff
@@ -330,6 +334,31 @@ def test_options_request(self):
             ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
         }
 
+    def test_cors_headers_param_request(self) -> None:
```
redlickigrzegorz, line 337: It would be also good to add a test that shows that headers from the parameter are more important than the ones from the env variable - as we saw, is easy to change this order.

redlickigrzegorz (approved):

PR #49, opened 2022-04-12
Adds Event Broker and iSort

pdyba:


examples/simple_resource.py
```diff
@@ -11,15 +11,15 @@
 from lbz.router import add_route
 
 
-class HelloWorld(Resource):
+class HelloWorldExample(Resource):
```
redlickigrzegorz, line 14: What was the problem with the previous name? It is used also in `README.md`, so I would change that in both places, if we really need 😉

setup.cfg
```diff
@@ -59,3 +59,15 @@ show_error_codes = True
...
 max-line-length = 99
+
+[isort]
+# https://pycqa.github.io/isort/docs/configuration/options
+combine_as_imports = True
+include_trailing_comma = True
+line_length = 99
+multi_line_output = 3
+reverse_relative = True
+;skip_gitignore = True TODO: enable it
```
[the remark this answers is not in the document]
redlickigrzegorz: I see now that the `make` is installed additionally in pipelines. Maybe we can add also `git`?

That might solve the problem, don't you think? 🤔

lbz/events/broker.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@dataclass()
+class Event:
+    type: str
+    data: dict
+
+
+class EventBroker:
+    def __init__(self, mapper: Dict[str, List[Callable]], raw_event: dict) -> None:
```
redlickigrzegorz, line 16: I liked the implementation in which `raw_event` was passed directly to the `handle()` method.

Thanks to that we will be able to handle multiple events using one object of this class - more natural 😉

lbz/events/broker.py
```diff
@@ -0,0 +1,43 @@
...
+        self.mapper = mapper
+        self.event = Event(type=raw_event["detail-type"], data=raw_event["detail"])
+
+    def handle(
+        self,
```
redlickigrzegorz, line 21: It looks strange formatted this way - put in one line 😅

lbz/events/broker.py
```diff
@@ -0,0 +1,43 @@
...
+            except Exception as error:  # noqa pylint: disable=broad-except, unused-variable
+                logger.exception("Handling event failed, event: %s", self.event)
+
+        self.post_handle()
+
+    def _get_handlers(self) -> List[Callable]:
+        try:
+            return self.mapper[self.event.type]
+        except KeyError as err:
+            raise NotImplementedError(f"Trigger {self.event.type} not implemented") from err
```
redlickigrzegorz, line 37: What would you say for: "_No handlers implemented for {self.event.type}_`?

It suits better to the naming you used in this module 😉

lbz/events/broker.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def handle(
+        self,
+    ) -> None:
+        self.pre_handle()
+
+        for handler in self._get_handlers():
+            try:
+                handler(self.event)
+            except Exception as error:  # noqa pylint: disable=broad-except, unused-variable
```
redlickigrzegorz, line 28: Don't create an `error` variable if you don't use it 😉 

Thanks to that, you will be able to remove skipping `unused-variable`.

tests/test_events_broker.py
```diff
@@ -0,0 +1,43 @@
+from unittest.mock import MagicMock
```
redlickigrzegorz, line 1: According to our convention that should be used also in this repository, these tests should be put in:
```
tests/test_events/test_broker.py
```

tests/test_events_broker.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pytest
+
+from lbz.events.broker import Event, EventBroker
+
+
+class TestMyEventBroker:
```
redlickigrzegorz, line 8: `TestEventBroker` - without `My`.

tests/test_events_broker.py
```diff
@@ -0,0 +1,43 @@
...
+        expected_event = Event(type="x", data={"y": 1})
+        mapper = {"x": [func_1, func_2]}
+        event = {"detail-type": "x", "detail": {"y": 1}}
+
+        EventBroker(mapper, event).handle()
+
+        func_1.assert_called_once_with(expected_event)
+        func_2.assert_called_once_with(expected_event)
+
+    def test_broker_raises_not_implemented_when_event_type_is_not_handled(self) -> None:
```
redlickigrzegorz, line 21: "_... when event type is not recognized_" that is what you test here 😉

tests/test_events_broker.py
```diff
@@ -0,0 +1,43 @@
...
+
+        with pytest.raises(NotImplementedError, match="Trigger y not implemented"):
+            EventBroker(mapper, event).handle()
+
+        func_1.assert_not_called()
+        func_2.assert_not_called()
+
+    def test_broker_continues_even_if_one_handler_failed(self) -> None:
+        func_1 = MagicMock()
+        func_2 = MagicMock(side_effect=TypeError)
```
redlickigrzegorz, line 35: It would be good to check that error has been logged during handling an event - I mean using `caplog` fixture.

redlickigrzegorz, line 35: Also, this is the last `handler`, so we cannot be 100% sure the broker continues the work after this failure.

You need to add something after this error to show it really works as expected 😉

examples/simple_resource.py
pdyba: pylint.

lbz/events/broker.py
pdyba: Lambda does not allow for such case ATM.
1 event = 1 invocation

pdyba: and above give you access to the even at all times.

tests/test_events_broker.py
pdyba: in LBZ we don't follow that convention yet.

examples/simple_resource.py
redlickigrzegorz: But `pylint` what? 🤔  Do you mean code duplication? And as I see caused by imports...

We should disable this warning for time being as we did in other places 😉

lbz/events/broker.py
redlickigrzegorz: I mentioned it from the code/Python perspective, not Lambda 😉 

And yes, there are benefits of both solutions, so definitely, I won't force changing that 👍

tests/test_events_broker.py
redlickigrzegorz: But we should and I asked for that in the previous pull requests 😉 

There won't be a better moment to start doing it better, especially since it requires ~1 minute of work.

CHANGELOG.md
```diff
@@ -189,4 +189,13 @@ Released 2021-10-09
...
\ No newline at end of file
+- Moves All JWT validation to decoding function from resource
+
+
+
+### Version 0.5.2
+Released 2022-04-12
+
+- Adds Event Broker to support Asynchronous event handling with AWS Lambda
+- Introduces iSort for better sorting and related checks
```
redlickigrzegorz, line 200: This is `isort`, not `iSort` - this way it looks like something from Apple 😃

tests/test_events_broker.py
```diff
@@ -0,0 +1,53 @@
...
+        mapper = {"x": [func_1, func_2]}
+        event = {"detail-type": "y", "detail": {"y": 1}}
+
+        with pytest.raises(NotImplementedError, match="No handlers implemented for y"):
+            EventBroker(mapper, event).handle()
+
+        func_1.assert_not_called()
+        func_2.assert_not_called()
+
+    def test_broker_continues_even_if_one_handler_failed(self, caplog) -> None:
```
redlickigrzegorz, line 34: You skipped type annotation for `caplog`.

redlickigrzegorz (approved):

PR #50, opened 2022-04-14
Adds EventApi and EventAware Resource + more

pdyba:
- Adds Response as a part of Resource so it can be easly access by post hook.
- Improves mypy situation for tests


lbz/boto3_client.py
```diff
@@ -0,0 +1,61 @@
+from functools import cached_property
```
redlickigrzegorz, line 1: Originally, this module was named simply `boto3_.py` and it was done this way to avoid repeating the `client` word too many times.

But if you don't like this convention, we can go with `boto3_broker.py` or `boto3_adapter.py` 😉

lbz/boto3_client.py
```diff
@@ -0,0 +1,61 @@
...
+    EventBridgeClient = object
+    LambdaClient = object
+    S3Client = object
+    SNSClient = object
+    SSMClient = object
+    SQSClient = object
+    DynamoDBClient = object
+
+
+class Boto3Clients:
```
redlickigrzegorz, line 27: This should be a singular form: `Boto3Client`.

lbz/events/api.py
```diff
@@ -0,0 +1,58 @@
+import json
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
redlickigrzegorz, line 4: Put such `if` conditions below all the normal imports, otherwise they are wrongly sorted 😉

tests/test_resource.py
```diff
@@ -414,20 +421,33 @@ def test_get_pagination_multifield_query_params(self):
...
-    def test_pagination_uri_without_query_params(self):
+    def test_pagination_uri_without_query_params(self) -> None:
         self.resource.request.query_params = MultiDict({})
         expected = "/test/path?offset={offset}&limit={limit}"
         assert self.resource._pagination_uri == expected  # pylint: disable=protected-access
+
+
+class TestEventAwareResource:
+    @patch.object(EventAPI, "send")
+    def test___call__(self, mocked_event_api_send: MagicMock) -> None:
```
redlickigrzegorz, line 445: I know why you put `__call__` into the name of this test but the name of the class suggests what is tested.

It would be much more helpful if you describe what behavior you test here instead 😉

tests/test_resource.py
```diff
@@ -414,20 +421,33 @@ def test_get_pagination_multifield_query_params(self):
...
+        class XResource(EventAwareResource):
+            @add_route("/")
+            def test_method(self):
+                return Response({"message": "x"})
+
+        XResource(event)()
+
+        mocked_event_api_send.assert_called_once()
```
redlickigrzegorz, line 453: Based on the above comments and a possible problem that was present even in the previous versions, I think, you should create additional tests for this logic, especially for the error situations 😉

lbz/events/api.py
```diff
@@ -0,0 +1,58 @@
...
+
+class BaseEvent:
+    type: str
+
+    def __init__(self, raw_data: dict) -> None:
+        self.raw_data = raw_data
+        self.data: str = self.serialize(raw_data)
+
+    @staticmethod
+    def serialize(data: dict) -> str:
```
redlickigrzegorz, line 24: Just to stay consistent in naming, I would put here `raw_data` as an argument.

lbz/events/api.py
```diff
@@ -0,0 +1,58 @@
...
+        return json.dumps(data, default=str)
+
+
+class EventAPI(metaclass=Singleton):
+    def __init__(self) -> None:
+        self._source = getenv("AWS_LAMBDA_FUNCTION_NAME") or "lbz-event-api"
+        self._resources: List[str] = []
+        self.entries: List[PutEventsRequestEntryTypeDef] = []
```
redlickigrzegorz, line 32: I think both `entries` and `bus_name` should be protected as well.

I don't see any good reason for having access to them from outside.

lbz/events/api.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def register(self, new_event: BaseEvent) -> None:
+        self.entries.append(self.create_eb_entry(new_event))
+
+    def get_all_registered_events(self) -> List[PutEventsRequestEntryTypeDef]:
+        return self.entries
+
+    def send(self) -> None:
+        if self.entries:
+            clients.eventbridge.put_events(Entries=self.entries)
```
redlickigrzegorz, line 49: Just to be 100% sure it won't duplicate events, you should make `self.entries` empty after this operation 😉

lbz/events/api.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, raw_data: dict) -> None:
+        self.raw_data = raw_data
+        self.data: str = self.serialize(raw_data)
+
+    @staticmethod
+    def serialize(data: dict) -> str:
+        return json.dumps(data, default=str)
+
+
+class EventAPI(metaclass=Singleton):
```
redlickigrzegorz, line 28: I think it might be named `EventBridge` or `EventBus`, don't you think? Just a suggestion as API here might be misunderstood.

lbz/resource.py
```diff
@@ -230,3 +231,9 @@ def _pagination_uri(self) -> str:
             encoded_params = urlencode(query_params, doseq=True)  # type: ignore
             return f"{self.urn}?{encoded_params}&offset={{offset}}&limit={{limit}}"
         return f"{self.urn}?offset={{offset}}&limit={{limit}}"
+
+
+class EventAwareResource(Resource):
+    def post_request_hook(self) -> None:
+        if self.response.is_ok():
+            EventAPI().send()
```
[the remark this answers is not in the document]
pdyba: its resolved alongs side the original issue.

CHANGELOG.md
```diff
@@ -192,10 +192,22 @@ Released 2021-10-09
 - Moves All JWT validation to decoding function from resource
 
 
+### Version 0.5.1
```
redlickigrzegorz, line 195: You skipped the release date: `2021-11-17`

CHANGELOG.md
```diff
@@ -192,10 +192,22 @@ Released 2021-10-09
...
+- Updates dependencies
+
+
+### Version 0.5.3
+Released 2022-04-15
+
+- Adds EventApi and EventAware Resource
+- Adds Response as a part of Resource, so it can be easily accessed by post hook.
+- Improves mypy situation for tests
+- Makes Resource's post_request_hook error proneŻ
```
redlickigrzegorz, line 213: Typo: "_prone**Ż**_" but rather you wanted to write "_error-proof_" or "_fault-tolerant_"

tests/test_resource.py
```diff
@@ -414,20 +421,54 @@ def test_get_pagination_multifield_query_params(self):
...
 
-    def test_pagination_uri_without_query_params(self):
+    def test_pagination_uri_without_query_params(self) -> None:
         self.resource.request.query_params = MultiDict({})
         expected = "/test/path?offset={offset}&limit={limit}"
         assert self.resource._pagination_uri == expected  # pylint: disable=protected-access
+
+
+class TestEventAwareResource:
+    def test___init__(self) -> None:
```
redlickigrzegorz, line 444: You did exactly the same thing as with the [\_\_call\_\_()](https://github.com/pdyba/lambdalizator/pull/50#discussion_r850446107) before. The name of this test brings almost no value when you don't mention what you expect.

tests/test_resource.py
```diff
@@ -414,20 +421,54 @@ def test_get_pagination_multifield_query_params(self):
...
+        class XResource(EventAwareResource):
+            @add_route("/")
+            def test_method(self) -> Response:
+                return Response({"message": "x"})
+
+        res = XResource(event)
+        assert hasattr(res, "event_api")
+        assert res.event_api == EventAPI()
```
redlickigrzegorz, line 452: Shouldn't you use `is` to compare them? This should be exactly the same object, shouldn't it?

tests/test_resource.py
```diff
@@ -414,20 +421,54 @@ def test_get_pagination_multifield_query_params(self):
...
+            @add_route("/")
+            def test_method(self) -> Response:
+                return Response({"message": "x"})
+
+        XResource(event)()
+
+        mocked_event_api_send.assert_called_once()
+
+    @patch.object(EventAPI, "send")
+    def test_post_hook_event_api_did_not_sent(self, mocked_event_api_send: MagicMock) -> None:
```
redlickigrzegorz, line 466: It would be good to put an explanation of why it is not sent - simpler to understand in the future instead of reading implementation each time 😉

lbz/resource.py
```diff
@@ -71,19 +73,18 @@ def __call__(self) -> Response:
...
                 logger.warning(err, exc_info=is_in_debug_mode())
-            return err.get_response(self.request.context["requestId"])
+            self.response = err.get_response(self.request.context["requestId"])
         except Exception as err:  # pylint: disable=broad-except
             logger.exception(err)
-            return ServerError().get_response(self.request.context["requestId"])
-        finally:
-            self.post_request_hook()
+            self.response = ServerError().get_response(self.request.context["requestId"])
+        self._post_request_hook()
```
redlickigrzegorz, line 86: You didn't create any test confirming that failure inside this function does not affect the response being delivered to the user.

This is a pretty important fix in this pull request and I cannot leave it as it is now 😉

lbz/events/api.py
redlickigrzegorz: As we agreed, we keep the current name, at least for now.

lbz/events/api.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def send(self) -> None:
+        if self._pending_events:
+            entries = [self._create_eb_entry(event) for event in self._pending_events]
+            client.eventbridge.put_events(Entries=entries)
+            self._mark_sent()
+
+    def _create_eb_entry(self, new_event: BaseEvent) -> PutEventsRequestEntryTypeDef:
+        """
+        boto3.amazonaws.com/v1/documentation/api/latest/reference/services/events.html#EventBridge.Client.put_events
```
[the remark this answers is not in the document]
redlickigrzegorz: But for me, it can be removed at all as everything is described by `boto3-stubs` - this is better to rely on it instead of putting links to documentation every time we use something from `boto3` 🙂

lbz/events/api.py
```diff
@@ -0,0 +1,75 @@
...
+        return self._pending_events
+
+    def get_all_sent_events(self) -> List[BaseEvent]:
+        return self._sent_events
+
+    def send(self) -> None:
+        if self._pending_events:
+            entries = [self._create_eb_entry(event) for event in self._pending_events]
+            client.eventbridge.put_events(Entries=entries)
+            self._mark_sent()
```
redlickigrzegorz, line 59: I realized that we should clear the list of pending events once an error appears, otherwise, in the worst scenario, they will wait for being sent and if it happens, it will be very confusing and hard to debug 😬 

We can introduce the list of `failed` events or something similar.

redlickigrzegorz: I started also thinking why it has to be Singleton - if we resign from that, this problem disappear 🤔

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+
+from lbz.aws_boto3 import Boto3Client
+from lbz.events.api import BaseEvent, EventAPI
+
+
+class MyTestEvent(BaseEvent):
+    type = "MY_TEST_EVENT"
+
+
+class TestBaseNewEvent:
```
redlickigrzegorz, line 13: `TestBaseEvent` - without `New`

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+    def test_base_new_event(self) -> None:
```
redlickigrzegorz, line 14: The initialization is tested multiple times below - I don't see value in testing it separately.

Much better would be to focus on tests that bring real value - I mentioned examples in the above comments 😉

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+
+        assert hasattr(new_event, "raw_data")
+        assert hasattr(new_event, "data")
+        assert new_event.type == "MY_TEST_EVENT"
+        assert new_event.raw_data == {"x": 1}
+        assert new_event.data == '{"x": 1}'
+
+
+class TestEventApi:
+    event_api: Optional[EventAPI] = None
```
redlickigrzegorz, line 25: Why do you need that here?

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+        with patch.dict(environ, {"AWS_LAMBDA_FUNCTION_NAME": "million-dollar-lambda"}):
+            self.event_api = EventAPI()
+
+    def teardown_method(self, _test_method) -> None:
+        # pylint: disable= attribute-defined-outside-init
+        self.event_api._del()  # pylint: disable=protected-access
+        self.event_api = None
+
+    def test_init_with_defaults(self) -> None:
+        # pylint: disable=protected-access
```
redlickigrzegorz, line 38: We discussed that multiple times, we shouldn't test protected attributes. They are protected on purpose and shouldn't be used outside the class...

Generally speaking, by these tests, you hardcoded the names of these attributes which is not a good habit.

redlickigrzegorz: Here you have [the first article I found about this topic](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html) if you think I am a crazy person touching this topic here 😅

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+        self.event_api.set_resources(["x", "Y"])
+        assert self.event_api._resources == ["x", "Y"]
+
+    def test_register(self) -> None:
+        # pylint: disable=protected-access
+        event = MyTestEvent({"x": 1})
+        self.event_api.register(event)
+
+        assert self.event_api._pending_events == [event]
```
redlickigrzegorz, line 60: You shouldn't use protected attributes this way as well.

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_get_all_registered_events(self) -> None:
+        event_1 = MyTestEvent({"x": 1})
+        event_2 = MyTestEvent({"x": 2})
+        self.event_api.register(event_1)
+        self.event_api.register(event_2)
+        assert self.event_api.get_all_pending_events() == [event_1, event_2]
+
+    @patch.object(Boto3Client, "eventbridge")
+    @patch.object(EventAPI, "_mark_sent")
```
redlickigrzegorz, line 70: This method is one of the simplest possible, why do you want to mock it? 😐

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+                    "DetailType": "MY_TEST_EVENT",
+                    "EventBusName": "million-dollar-lambda-event-bus",
+                    "Resources": [],
+                    "Source": "million-dollar-lambda",
+                }
+            ]
+        )
+        mock__mark_sent.assert_called_once()
+
+    def test_clear(self):
```
redlickigrzegorz, line 90: Besides the above comments, you should improve test naming. I know tests in `lbz` are not ideal but I don't want to make this situation worse when we write new code.

tests/test_events_api.py
```diff
@@ -0,0 +1,137 @@
...
+        assert resp == {
+            "Detail": '{"x": 1}',
+            "DetailType": "MY_TEST_EVENT",
+            "EventBusName": "million-dollar-lambda-event-bus",
+            "Resources": [],
+            "Source": "million-dollar-lambda",
+        }
+
+
+class TestEventApiSingleton:
```
redlickigrzegorz, line 128: This should be an integral part of the above class as you still test `EventAPI`.

redlickigrzegorz (changes requested):

tests/test_events_api.py
pdyba: mypy

redlickigrzegorz: But you removed this line anyway, so it stopped being a problem?

tests/test_events_api.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+        assert str(self.event_api) == expected_repr
+        self.event_api._mark_sent()  # pylint: disable=protected-access
+        expected_repr = (
+            "<EventAPI bus: million-dollar-lambda-event-bus Events: pending=0 sent=2 failed=0>"
+        )
+        assert str(self.event_api) == expected_repr
+        self.event_api.register(MyTestEvent({"x": 1}))
+        self.event_api.register(MyTestEvent({"x": 1}))
+        self.event_api._mark_failed()  # pylint: disable=protected-access
```
redlickigrzegorz, line 60: Ok, testing the representation might be useful but this test should be as simple as possible, definitely not containing three separate scenarios that base on the protected methods...

redlickigrzegorz: This class has two major public functions, `register()` and `send()` - test them deeply instead of testing everything around. Thanks to that you will probably achieve a shorter test module that will be more valuable in the future 😉

tests/test_events_api.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.event_api._mark_failed()  # pylint: disable=protected-access
+
+        assert self.event_api.get_all_failed_events() == [event]
+
+    @patch.object(Boto3Client, "eventbridge")
+    def test_send_no_events(self, mock_send: MagicMock) -> None:
+        self.event_api.send()
+
+        mock_send.assert_not_called()
```
redlickigrzegorz, line 149: This mock will never be called no matter what happens. You should check it this way:
```python
mock_send.put_events.assert_not_called()
```

tests/test_events_api.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.event_api._mark_sent()  # pylint: disable=protected-access
+
+        assert self.event_api.get_all_sent_events() == [event]
+
+    def test_get_all_failed_events(self) -> None:
+        event = MyTestEvent({"x": 1})
+        self.event_api.register(event)
+
+        self.event_api._mark_failed()  # pylint: disable=protected-access
```
redlickigrzegorz, line 141: In the previous comments, I mentioned that testing or using protected methods is not a good idea. Here you have an example of why: the test using a public method would catch that this one is not used anywhere 😉

redlickigrzegorz (changes requested):

lbz/events/api.py
```diff
@@ -0,0 +1,80 @@
...
+            "Detail": new_event.data,
+            "DetailType": new_event.type,
+            "EventBusName": self._bus_name,
+            "Resources": self._resources,
+            "Source": self._source,
+        }
+
+    def _mark_sent(self) -> None:
+        self._sent_events = self._pending_events
+        self._pending_events = []
```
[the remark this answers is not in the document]
pdyba: clearing that would make only sens if we have implemented retry logic that can be used by library user - we don't know how it was used actually maybe there were 2 or more send()'s in the process.

lbz/events/api.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _mark_failed(self) -> None:
+        self._failed_events = self._pending_events
+        self._pending_events = []
```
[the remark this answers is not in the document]
pdyba: as above that might be incorrect.

lbz/events/api.py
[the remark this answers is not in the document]
redlickigrzegorz: As we discussed, then the user can get a list of failed/succeeded events between sending without consequences 🙂

tests/test_events_api.py
```diff
@@ -0,0 +1,166 @@
...
+            ]
+        )
+
+    def test_register(self) -> None:
+        assert self.event_api.get_all_pending_events() == []
+
+        event = MyTestEvent({"x": 1})
+        self.event_api.register(event)
+
+        assert self.event_api.get_all_pending_events() == [event]
```
redlickigrzegorz, line 82: This test is exactly the same as the below one but tests the single event - I think we can remove this one 😉

tests/test_events_api.py
```diff
@@ -0,0 +1,166 @@
...
+    @patch.object(Boto3Client, "eventbridge", MagicMock())
+    def test_get_all_sent_events(self) -> None:
+        assert self.event_api.get_all_sent_events() == []
+
+        event = MyTestEvent({"x": 1})
+        self.event_api.register(event)
+
+        self.event_api.send()
+
+        assert self.event_api.get_all_sent_events() == [event]
```
redlickigrzegorz, line 122: This is tested the same way above, I don't think we need to duplicate it.

tests/test_events_api.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @patch.object(Boto3Client, "eventbridge")
+    def test_get_all_failed_events(self, mock_send: MagicMock) -> None:
```
redlickigrzegorz, line 125: It would be good to say that this test checks the situation when sending events failed: the `send()` method is the most important here.

tests/test_events_api.py
```diff
@@ -0,0 +1,166 @@
...
+        with pytest.raises(NotADirectoryError):
+            self.event_api.send()
+
+        assert self.event_api.get_all_failed_events() == [event]
+
+    @patch.object(Boto3Client, "eventbridge")
+    def test_send_no_events(self, mock_send: MagicMock) -> None:
+        self.event_api.send()
+
+        mock_send.put_events.assert_not_called()
```
redlickigrzegorz, line 141: You can also check that all the lists are empty, just to present the state more directly.

redlickigrzegorz (approved):

tests/test_events_api.py
```diff
@@ -0,0 +1,166 @@
...
+    def setup_method(self) -> None:
+        # pylint: disable= attribute-defined-outside-init
+        with patch.dict(environ, {"AWS_LAMBDA_FUNCTION_NAME": "million-dollar-lambda"}):
+            self.event_api = EventAPI()
+
+    def teardown_method(self, _test_method: Callable) -> None:
+        # pylint: disable= attribute-defined-outside-init
+        self.event_api._del()  # type: ignore # pylint: disable=protected-access
+
+    @patch.object(Boto3Client, "eventbridge", MagicMock())
```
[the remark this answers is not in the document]
pdyba: I thought about it and multiple patches looked cleaner - more readable.

lbz/events/api.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_all_failed_events(self) -> List[BaseEvent]:
+        return self._failed_events
+
+    def send(self) -> None:
+        try:
+            if self._pending_events:
+                entries = [self._create_eb_entry(event) for event in self._pending_events]
+                client.eventbridge.put_events(Entries=entries)
+                self._mark_sent()
```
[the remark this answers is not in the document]
pdyba: why would You want to trigger a function that should do nothing (unless there were failed events in 'saved')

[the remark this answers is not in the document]
redlickigrzegorz: For the same reason as in all the other edge cases, the state after triggering `send()` should show truth instead of something in-between 😉

PR #53, opened 2022-06-03
Always send events in chunks to stay within AWS limits

redlickigrzegorz:
During the first real usage of the newly implemented EventAPI, it turned out that sending more events at once is impossible with the current implementation. That is caused by [AWS limits](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html):
> Array Members: Minimum number of 1 item. Maximum number of 10 items.

Here, in this pull request, I have implemented the fix that slice the list of pending events into the chunks and invokes the AWS function multiple times to deliver all of them.

---
Besides the main fix, I have gone through the repository files and upgraded them to the latest version used in other repositories - mostly, to improve the work on this library 😉


Makefile
```diff
@@ -1,34 +1,88 @@
...
 
-build:
-	python setup.py sdist bdist_wheel
-
-upload:
-	python3 -m twine upload dist/*
+###############################################################################
+# Python Requirements
+# -----------------------------------------------------------------------------
+# TODO: Start using pip-tools as a dependency freezing tool
```
redlickigrzegorz, line 9: _FYI:_ This pull request is quite big without `pip-tools` I didn't want to go too far with changes 😉

lbz/events/api.py
```diff
@@ -66,8 +69,9 @@ def get_all_failed_events(self) -> List[BaseEvent]:
 
     def send(self) -> None:
         try:
-            if self._pending_events:
-                entries = [self._create_eb_entry(event) for event in self._pending_events]
+            for i in range(0, len(self._pending_events), MAX_EVENTS_TO_SEND_AT_ONCE):
```
redlickigrzegorz, line 72: _FYI:_ This is the real fix I have implemented, the rest of the changes are related to the repository files.

setup.cfg
```diff
@@ -1,7 +1,57 @@
+[flake8]
```
redlickigrzegorz, line 1: _FYI:_ I have fixed the order of configs (alphabetical) - thanks to that, it will be easier to keep that up to date with other repositories 😉

lbz/events/api.py
pdyba: with the try except logic it should _mark_sent for those sent in each try.
also I would move then try to within for loop.
so both mark sent and mark field should take chunks and mark only them the for loop should become a while as the pending_events will be mutating (for will fail) 
consider adding _get_next_chunk_of_events method that will utilise that limit.

redlickigrzegorz: I completely agree with you, I don't know why I skipped this part 🤦

lbz/events/api.py
```diff
@@ -52,9 +56,22 @@ def set_resources(self, resources: List[str]) -> None:
...
+        return deepcopy(self._pending_events)
+
+    @property
+    def failed_events(self) -> List[BaseEvent]:
+        return deepcopy(self._failed_events)
+
     def register(self, new_event: BaseEvent) -> None:
         self._pending_events.append(new_event)
 
+    # TODO: Stop sharing protected lists outside the class, use the above properties instead
```
redlickigrzegorz, line 74: _FYI:_ I have added above the properties that always return a copy of events, just to do it the right way.

If you think, we can remove the below methods, I can do this but I left them for backward compatibility 😉

lbz/events/api.py
```diff
@@ -87,13 +110,3 @@ def _create_eb_entry(self, new_event: BaseEvent) -> PutEventsRequestEntryTypeDef
             "Resources": self._resources,
             "Source": self._source,
         }
-
-    def _mark_sent(self) -> None:
```
redlickigrzegorz, line 91: _FYI:_ After the last changes, both methods became single lines that were used in one place - did not make much sense to keep them.

lbz/events/api.py
```diff
@@ -65,13 +82,19 @@ def get_all_failed_events(self) -> List[BaseEvent]:
...
 
     def send(self) -> None:
+        self._sent_events = []
+        self._failed_events = []
+
         try:
-            if self._pending_events:
-                entries = [self._create_eb_entry(event) for event in self._pending_events]
+            while self._pending_events:
+                events = self._pending_events[:MAX_EVENTS_TO_SEND_AT_ONCE]
```
redlickigrzegorz, line 90: _FYI:_ Now, everything is in one method which looks self-descriptive to me, but if you think it is too long or something, I can move this part to a separate one 😉

lbz/events/api.py
```diff
@@ -65,13 +82,20 @@ def get_all_failed_events(self) -> List[BaseEvent]:
...
+                entries = [self._create_eb_entry(event) for event in events]
                 client.eventbridge.put_events(Entries=entries)
-            self._mark_sent()
+
+                self._sent_events.extend(events)
+                self._pending_events = self._pending_events[MAX_EVENTS_TO_SEND_AT_ONCE:]
         except Exception as err:
-            self._mark_failed()
+            self._failed_events.extend(self._pending_events)
+            self._pending_events = []
```
pdyba, line 98: Why would You empty that list if the events were not send and they possible could be?
That would make debugging harder and b. that would not allow tu build retry logic around that method. the line 
`self._pending_events = self._pending_events[MAX_EVENTS_TO_SEND_AT_ONCE:]`
should be used here and then I would add a flag either on Init level or on method level `raise_at_first_sending_failure` +/- and only if that flag is present do actual raise on first failure in other case send remaining and not raise we could also try doing it a 3 way switch with
a) raise at first (and the the current logic makes more sense)
b) raise after all if there was an error 
c) never raise if anything was send (raise only when nothing was send and something should be)

tests/test_events/test_api.py
```diff
@@ -122,6 +137,36 @@ def test_send(self, mock_send: MagicMock) -> None:
...
 
+    @patch.object(Boto3Client, "eventbridge")
+    def test__send__sends_events_in_chunks_respecting_limits(self, mock_send: MagicMock) -> None:
+        for i in range(33):  # AWS allows sending maximum 10 events at once
+            self.event_api.register(MyTestEvent({"x": i}))
+
+        self.event_api.send()
+
+        assert mock_send.put_events.call_count == 4
+        assert len(mock_send.put_events.call_args_list[0].kwargs["Entries"]) == 10
```
pdyba, line 148: You could to it with has calls and use array multiplication so it would look like that ~:
 == ([MyTestEvent({"x": i})] * 10, [MyTestEvent({"x": i})] * 10, [MyTestEvent({"x": i})] * 3,) 
 as Event has __eq__ implement :)

pdyba (changes requested):

lbz/events/api.py
redlickigrzegorz: So, this implementation is the **A)** scenario, and I empty this list here only because all of its events are marked as failed above. I did it this way to stay compatible with our previous agreements that the `send()` method does not leave anything pending 😉 

Also, our previous agreements touched the retrying logic - we decided it should be implemented outside this library if anyone needs it, and with this code, still, it is possible, based on the `event_api.failed_events` list. 

Debugging shouldn't be a problem as well, as still, the error is raised, nothing will be hidden 😉

redlickigrzegorz: I was considering the **B)** scenario for a moment but I thought that getting multiple errors from a single function may produce more unnecessary chaos. But we can talk about it and go this way 👍 

But I would be against giving the possibility to choose because we talk about the situations that should never happen, and if happen, they need to be fixed as soon as possible 😉

tests/test_events/test_api.py
redlickigrzegorz: That would be true if we test here events, but this line is about entries - they are quite big dictionaries.

I did it this way because I wanted to keep it clean - the main goal is to check whether this method was triggered 4 times with a max of 10 entries per run. The correctness of building the entries is tested in other tests 😉

lbz/events/api.py
```diff
@@ -65,14 +84,23 @@ def get_all_failed_events(self) -> List[BaseEvent]:
...
-            raise err
+                self._sent_events.extend(events)
+            except Exception as err:  # pylint: disable=broad-except
+                self._failed_events.extend(events)
+                logger.exception(err)
+
+            self._pending_events = self._pending_events[MAX_EVENTS_TO_SEND_AT_ONCE:]
+
+        if self._failed_events:
+            raise RuntimeError("Sending events has failed. Check logs for more details!")
```
redlickigrzegorz, line 103: _FYI:_ I was considering creating a custom Exception for this situation but it looks that lbz is raising mostly HTTP-like errors or `RuntimeError`, so I decided to stay consistent 😉 

This way or another, each error will be logged individually by the 98. line.

lbz/exceptions.py
```diff
@@ -1,7 +1,5 @@
 # coding=utf-8
-"""
-Set of HTTP Exceptions that are json compatabile.
```
redlickigrzegorz, line 3: _FYI:_ I have removed this docstring as this does not conform to all exceptions being declared here.

tests/test_events/test_api.py
```diff
@@ -122,6 +139,48 @@ def test_send(self, mock_send: MagicMock) -> None:
...
+
+    @patch.object(Boto3Client, "eventbridge")
+    def test__send__always_tries_to_send_all_events_treating_each_chunk_individually(
+        self, mock_send: MagicMock, caplog: LogCaptureFixture
+    ) -> None:
+        mock_send.put_events.side_effect = (
+            None,  # no error == success
+            ValueError("Event data is too big to be sent"),
+            None,  # no error == success
+            ValueError("Event type cannot be recognized"),
```
redlickigrzegorz, line 166: _FYI:_ These errors do not matter that much but I put here message to show logs more correctly 😉

lbz/events/api.py
redlickigrzegorz: As we agreed outside this PR, I have changed the strategy to always try to send all registered events 🙂

pdyba (approved):

PR #54, opened 2022-07-23
Create a decorator that makes the function an event emitter

redlickigrzegorz:
In this pull request, I have added a new decorator that is able to make the function an event emitter. Meaning the function will emit all the pending events if only finishes with success, otherwise will clean the event queues 🙂 

_A very similar logic we have implemented in the [EventAwareResource](https://github.com/pdyba/lambdalizator/blob/1d785bb0f3c9b54898321428337df565c6ea6507/lbz/resource.py#L247) class._


lbz/events/api.py
```diff
@@ -115,3 +116,19 @@ def _create_eb_entry(self, new_event: BaseEvent) -> PutEventsRequestEntryTypeDef
             "Resources": self._resources,
             "Source": self._source,
         }
+
+
+def event_emitter(function: Callable) -> Callable:
```
redlickigrzegorz, line 121: Feel free to suggest a better name 😉

lbz/events/api.py
```diff
@@ -115,3 +116,19 @@ def _create_eb_entry(self, new_event: BaseEvent) -> PutEventsRequestEntryTypeDef
...
+    """Decorator that makes function an emitter - automatically sends pending events on success"""
+
+    @wraps(function)
+    def wrapped(*args: Any, **kwargs: Any) -> Any:
```
redlickigrzegorz, line 125: As I mentioned in the description, the same logic we have implemented in the [EventAwareResource](https://github.com/pdyba/lambdalizator/blob/1d785bb0f3c9b54898321428337df565c6ea6507/lbz/resource.py#L247) class but there we depend on the response status code.

pdyba (approved):

PR #59, opened 2022-08-01
Adds 425 Too Early exception

pdyba:


CHANGELOG.md
```diff
@@ -250,3 +257,10 @@ Released 2022-07-31
 
 - Logs an error when required ref is missing in the Authorization policy
 - Accepts the error_code also during initialization of the Exception
+
+
+### Version 0.5.11
```
redlickigrzegorz, line 262: You need to also update the `version` file itself.

lbz/exceptions.py
```diff
@@ -216,6 +216,13 @@ class FailedDependency(LambdaFWClientException):
     status_code = HTTPStatus.FAILED_DEPENDENCY.value
 
 
+class TooEarly(LambdaFWClientException):
```
redlickigrzegorz, line 219: You should also add this new Exception to the test module, to verify it the same way as others 😉 

In the future, it would be good to make it more automatic because it is simple to forget about it.

redlickigrzegorz (approved):

lbz/exceptions.py
pdyba: I actually thought its already automatic.

redlickigrzegorz: That's how you implemented it 🤷

PR #61, opened 2022-08-06
Makes sure that event broker is always passing original event and oth…

pdyba:
…er handlers cannot affect it.

Makes DevServer threaded, so it can be run in background, useful for testing


tests/test_dev_server.py
```diff
@@ -1,6 +1,8 @@
 # coding=utf-8
 import io
+import json
 import socket
+import urllib
```
redlickigrzegorz, line 5: I think you should import `urllib.request`, otherwise, PyCharm has problems finding the correct code (most likely `mypy` will have the same problems).

tests/test_dev_server.py
```diff
@@ -12,7 +14,7 @@
 
 class HelloWorld(Resource):
     @add_route("/", method="GET")
-    def list(self) -> Response:
+    def handler(self) -> Response:
```
redlickigrzegorz, line 17: `handler()` is not the best name when we have `get()` declared below 😉 

I would leave the previous version as it harms no one but if you want to improve these endpoints, apply changes to the whole `HelloWorld` class.

tests/test_dev_server.py
```diff
@@ -54,3 +56,19 @@ def test_my_dev_server() -> None:
     assert dev_serv.port == 8000
     assert dev_serv.address == "localhost"
     assert issubclass(dev_serv.my_handler, MyLambdaDevHandler)
+
+
+def test_server_can_run_in_background() -> None:
+    http_process = MyDevServer(HelloWorld, port=9999)
```
redlickigrzegorz, line 62: I would keep the naming convention from above: `dev_serv`

tests/test_dev_server.py
```diff
@@ -54,3 +56,19 @@ def test_my_dev_server() -> None:
...
+    http_process.start()
+    url = "http://localhost:9999/"
+
+    request = urllib.request.Request(url, method="GET")
+    try:
+        with urllib.request.urlopen(request) as response:
+            assert response.status == 200
+            assert json.loads(response.read().decode()) == {"message": "HelloWorld"}
+    except Exception as err:
+        raise err
```
redlickigrzegorz, line 72: You don't have to catch an Exception if the only thing you want to do is to raise it again 😉 

`finally` will work anyway.

tests/test_events/test_broker.py
```diff
@@ -68,3 +69,15 @@ def test_broker_continues_even_if_one_handler_failed(self, caplog: LogCaptureFix
                 "Handling event failed, event: Event(type='x', data={'y': 1})",
             )
         ]
+
+    def test_broker_events_are_not_affected_by_specific_event_handler(self) -> None:
+        def handler(event: Event) -> None:
+            expected_event = Event({"y": 1}, event_type="x")
+            assert event == expected_event
```
redlickigrzegorz, line 76: In general, this test is not bad but each error raised directly in the handler is caught and logged.

Try to run it without `deepcopy()` - the test will pass anyway 😉

redlickigrzegorz (changes requested):

lbz/dev/server.py
```diff
@@ -147,24 +148,30 @@ def do_OPTIONS(self) -> None:  # pylint: disable=invalid-name
...
     def __init__(self, acls: Type[Resource], address: str = "localhost", port: int = 8000):
         class MyClassLambdaDevHandler(MyLambdaDevHandler):
             cls: Type[Resource] = acls
 
+        super().__init__()
         self.my_handler = MyClassLambdaDevHandler
         self.address = address
         self.port = port
         self.server_address = (self.address, self.port)
+        self.httpd = ThreadingHTTPServer(self.server_address, self.my_handler)
```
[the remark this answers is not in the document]
pdyba: I've added more verbose information on what is happening.
That line only binds to the port not starts the server.

lbz/dev/server.py
```diff
@@ -147,24 +148,42 @@ def do_OPTIONS(self) -> None:  # pylint: disable=invalid-name
...
         """
         print(f"serving on http://{self.address}:{self.port}")
-        httpd = HTTPServer(self.server_address, self.my_handler)
-        httpd.serve_forever()
+        self.httpd.serve_forever()
+
+    def stop(self) -> None:
+        self.httpd.shutdown()
+        self.httpd.server_close()
+        print(f"Server stopped and port {self.port} release")
```
redlickigrzegorz, line 183: Typo: "_release**d**_"

tests/conftest.py
```diff
@@ -59,8 +59,8 @@ def sample_event() -> APIGatewayEvent:
     )
 
 
-@pytest.fixture(scope="session")
-def jwt_partial_payload() -> dict:
+@pytest.fixture(scope="session", name="jwt_partial_payload")
+def jwt_partial_authz() -> dict:
```
redlickigrzegorz, line 63: Some time ago we agreed to add the `_fixture` suffix to avoid `pylint` warnings.

Changing the name of the fixture function is not recommended as you introduce a new name.

tests/conftest.py
```diff
@@ -69,8 +69,10 @@ def jwt_partial_payload() -> dict:
     }
 
 
-@pytest.fixture(scope="session")
-def full_access_authz_payload(jwt_partial_payload) -> dict:  # pylint: disable=redefined-outer-name
+@pytest.fixture(scope="session", name="full_access_authz_payload")
+def full_access_authz(
+    jwt_partial_payload: dict,
```
redlickigrzegorz, line 74: Why can't it be placed in one line? To be honest, I prefer one line of declaration instead of three 😉

tests/conftest.py
```diff
@@ -102,13 +104,13 @@ def limited_access_auth_header(
     )
 
 
-@pytest.fixture(scope="session")
-def username() -> str:
+@pytest.fixture(scope="session", name="username")
+def username_generator() -> str:
```
redlickigrzegorz, line 108: The same here, now, it looks like can be used as a standalone function but it is a fixture 🤷

tests/conftest.py
```diff
@@ -123,17 +125,17 @@ def user_cognito(username, jwt_partial_payload) -> dict:  # pylint: disable=rede
 
 
 @pytest.fixture(scope="session")
-def user_token(user_cognito) -> str:  # pylint: disable=redefined-outer-name
+def user_token(user_cognito: dict) -> str:  # pylint: disable=redefined-outer-name
```
redlickigrzegorz, line 128: Why didn't you rename the `user_cognito` fixture? It would be good to remove these warnings as well 😉

tests/conftest.py
```diff
@@ -165,3 +167,19 @@ def garbage(self, restrictions) -> Response:  # pylint: disable=unused-argument
             return Response("x")
 
     return XResource
+
+
+@pytest.fixture()
+def sample_resource_without_authorization() -> Type[Resource]:
```
redlickigrzegorz, line 173: This can be named simply `sample_resource` if we have version `with_authorization` above 😉

redlickigrzegorz (approved):

PR #62, opened 2022-08-08
General Tests improvements and mypy improvements

pdyba:
- Adds Router cleanup after each test and router fixture
- General tests improvements
- Many mypy related improvements for tests


tests/conftest.py
```diff
@@ -26,11 +26,24 @@ def allowed_audiences() -> List[str]:
 
 
 @pytest.fixture(autouse=True)
-def clear_authz_collector():
+def clear_authz_collector() -> Generator:
```
redlickigrzegorz, line 29: It would be better to be more precise: `Generator[None, None, None]`.

Python allows using `Iterator[None]` if you want to get rid of multiple `None` 😉

tests/conftest.py
```diff
@@ -26,11 +26,24 @@ def allowed_audiences() -> List[str]:
...
+@pytest.fixture(autouse=True)
+def clear_router_collector() -> Generator:
+    yield
+    Router()._clear()  # pylint: disable=protected-access
+
+
+@pytest.fixture()
+def sample_router() -> Router:
+    router = Router()
+    router.add_route("/", "GET", "x")
```
redlickigrzegorz, line 43: I am a little bit surprised that `handler` is a string 😅 

I think I need to check this code more deeply when I find a moment.

tests/conftest.py
```diff
@@ -26,11 +26,24 @@ def allowed_audiences() -> List[str]:
...
 
 
+@pytest.fixture(autouse=True)
+def clear_router_collector() -> Generator:
+    yield
+    Router()._clear()  # pylint: disable=protected-access
+
+
+@pytest.fixture()
+def sample_router() -> Router:
```
redlickigrzegorz, line 41: If this fixture is used only in one test module, it is recommended to put it directly there 😉 

Otherwise, we will have a pretty big `conftest.py` which will be hard to maintain.

tests/conftest.py
```diff
@@ -181,3 +194,25 @@ def garbage(self, restrictions: dict) -> Response:  # pylint: disable=unused-arg
             return Response("x")
 
     return XResource
+
+
+@pytest.fixture()
+def sample_guest_resource() -> Type[Resource]:
```
redlickigrzegorz, line 200: The same about this fixture - you used it only in `test_authz_decorator.py`.

tests/conftest.py
```diff
@@ -181,3 +194,25 @@ def garbage(self, restrictions: dict) -> Response:  # pylint: disable=unused-arg
...
+    class GuestResource(Resource):  # pylint: disable=unused-variable
```
redlickigrzegorz, line 201: Do you need to disable `pylint` here?

tests/test_authz_utils.py
```diff
@@ -26,14 +28,14 @@ def test_check_permission_raises(
             check_permission(res_instance, "garbage")
 
     def test_check_permission_unauthorised(
-        self, sample_event, sample_resource_with_authorization
+        self, sample_event: dict, sample_resource_with_authorization: Type[Resource]
```
redlickigrzegorz, line 31: `sample_event: APIGatewayEvent` - thanks to that we will be more precise.

tests/test_router.py
```diff
@@ -6,48 +6,39 @@
 
 
 class TestRouter:
-    def setup_method(self) -> None:
-        # pylint: disable= attribute-defined-outside-init
-        self.router = Router()
-        self.router.add_route("/", "GET", "x")
+    def test___init__(self, sample_router: Router) -> None:
```
redlickigrzegorz, line 9: I would add TODO somewhere here that we should improve the router tests.

e.g. this one should not exist 😉

tests/conftest.py
```diff
@@ -26,11 +26,24 @@ def allowed_audiences() -> List[str]:
...
-def clear_authz_collector():
+def clear_authz_collector() -> Generator:
     yield
     authz_collector.clean()
 
 
+@pytest.fixture(autouse=True)
+def clear_router_collector() -> Generator:
+    yield
+    Router()._clear()  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
pdyba: But that's the point its only for tests.

redlickigrzegorz (approved):

PR #63, opened 2022-08-22
Config management

pdyba:


lbz/configuration/configs.py
```diff
@@ -0,0 +1,90 @@
...
+            val = self.getter()
+            val = val if val is not None else self.default
+            if val is None:
+                self._value = None
+            else:
+                self._value = self.parser(val) if self.parser else val
+        return self._value
+
+
+class KeyValueConfig(BaseConfig):
```
redlickigrzegorz, line 36: This can be simply replaced by a globally declared variable - at least I would do that this way to simplify declaring/usage 😉

lbz/configuration/configs.py
```diff
@@ -0,0 +1,90 @@
...
+class SSMConfig(BaseConfig):
+    def __init__(self, key: str, ssm_key: str, parser: Any = str, default: Optional[Any] = None):
+        super().__init__(key, parser=parser, default=default)
+        self.ssm_key = ssm_key
+
+    def getter(self) -> Any:
+        return SSM.get_parameter(self.ssm_key)
+
+
+class JSONFileConfig(BaseConfig):
```
redlickigrzegorz, line 75: Are we sure we need to support this? We are talking about the code that works in Lambda Function, so, I am not sure who and why would like to put the configuration in the JSON file 🤔 

Maybe I am wrong but as long as we don't have such a request, it seems to be not needed feature.

lbz/configuration/configs.py
```diff
@@ -0,0 +1,90 @@
...
+        json_key: Optional[str] = None,
+        parser: Any = str,
+        default: Optional[Any] = None,
+    ):
+        super().__init__(key, parser=parser, default=default)
+        self.json_file = json_file
+        self.json_key = json_key
+
+    def getter(self) -> Any:
+        with open(self.json_file, encoding="UTF-8") as file:
```
redlickigrzegorz, line 89: Also, the whole concept reads one config value at a time, so in this case, if someone has a JSON file with 30 configuration variables, the file will be read 30 times at the start 😬

lbz/configuration/managers.py
```diff
@@ -0,0 +1,36 @@
...
+        self.configs: Dict[str, BaseConfig] = {cfg.key: cfg for cfg in configs}
+        self.loaded_configs: Dict[str, Any] = {}
+        for cfg in configs:
+            value = cfg.get_value()
+            if value is cfg.default is None:
+                raise ConfigurationMissingKey(cfg.key)
+            self.loaded_configs[cfg.key] = value
+
+    def __getattr__(self, item: str) -> Any:
+        return self.loaded_configs[item]
```
redlickigrzegorz, line 18: Now, I understand why you mentioned the need of having `.pyi` modules 😉 

I need to say that this manager won't be easy to use/maintain, so, I think we should consider a replacement if this is intended to become a core logic in every single other repository 😬

lbz/configuration/managers.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class LazyLoadConfigManager:
```
redlickigrzegorz, line 21: What would you say if we start storing configuration this way (prototype):
```python
class Config:
    hardcoded_value: int = 123
    pre_loaded_env_value: str = os.environ["..."]

    @cached_property
    def lazy_loaded_env_value(self) -> str:
        return os.environ["..."]

    @cached_property
    def lazy_loaded_ssm_value(self) -> str:
        return aws_ssm.SSM.get_parameter("...")
```

PR #68, opened 2022-09-28
adds base handler and lambda api

pdyba:


lbz/lambdas/__init__.py
```diff
@@ -0,0 +1,4 @@
+from lbz.lambdas.broker import LambdaBroker
+from lbz.lambdas.client import LambdaClient
+from lbz.lambdas.enums import LambdaResult
```
redlickigrzegorz, line 3: Here, there will be also the `LambdaSource` class - just not to forget 😉

lbz/lambdas/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+from lbz.lambdas.exceptions import LambdaError
```
redlickigrzegorz, line 4: But also, you skipped the `lbz.lambdas.response` module.

lbz/lambdas/exceptions.py
```diff
@@ -0,0 +1,12 @@
+class LambdaError(Exception):
+    def __init__(
+        self,
+        function_name: str,
+        op: str,
+        result: str,
+        response: dict,
+    ) -> None:
```
redlickigrzegorz, line 8:
```suggestion
    def __init__(self, function_name: str, op: str, result: str, response: dict) -> None:
```

lbz/lambdas/response.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class LambdaResponse(TypedDict, total=False):
+    result: str
+    data: Optional[Any]
+    message: Optional[str]
+    error_code: Optional[str]
+
+
+def lambda_ok_response(data: dict = None) -> LambdaResponse:
```
redlickigrzegorz, line 13: [#977464631](https://github.com/pdyba/lambdalizator/pull/64#discussion_r977464631):
> `data` is not necessarily a dictionary - the whole response is, so we can put here whatever we want 🤔

lbz/lambdas/response.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, Optional, TypedDict
+
+from lbz.lambdas.enums import LambdaResult
+
+
+class LambdaResponse(TypedDict, total=False):
+    result: str
+    data: Optional[Any]
+    message: Optional[str]
```
redlickigrzegorz, line 9: [#981467399](https://github.com/pdyba/lambdalizator/pull/64#discussion_r981467399):
> `Optional` here does not refer to the presence of a specific key - it just means that the value may be `None`.
> 
> Do we want to allow passing `None` as the message? 🤔

lbz/lambdas/response.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, Optional, TypedDict
+
+from lbz.lambdas.enums import LambdaResult
+
+
+class LambdaResponse(TypedDict, total=False):
+    result: str
+    data: Optional[Any]
```
redlickigrzegorz, line 8: [#981468874](https://github.com/pdyba/lambdalizator/pull/64#discussion_r981468874):
> Also here, `Any` means anything (including `None`) 😉

lbz/lambdas/response.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def lambda_ok_response(data: dict = None) -> LambdaResponse:
+    resp = LambdaResponse(result=LambdaResult.OK)
+    if data is not None:
+        resp["data"] = data
+    return resp
+
+
+def lambda_error_response(
```
redlickigrzegorz, line 20: In the original code, we have tests for this simple function 😉

lbz/lambdas/response.py
redlickigrzegorz, line 13: It would be good to write tests for this function as well, as it contains more logic inside 😉

lbz/lambdas/client.py
```diff
@@ -0,0 +1,76 @@
...
+    def invoke(
+        cls,
+        function_name: str,
+        op: str,
+        data: Optional[dict] = None,
+        *,
+        allowed_error_results: Iterable[str] = None,
+        raise_if_error_resp: bool = False,
+        asynchronous: bool = False,
+    ) -> Union[InvocationResponseTypeDef, dict]:
```
redlickigrzegorz, line 39: [#965935825](https://github.com/pdyba/lambdalizator/pull/64#discussion_r965935825):
> This way of declaring return type gives you almost no benefits as the return value can always be a simple dictionary (there is no guarantee for `mypy` that keys are present inside).

lbz/lambdas/client.py
```diff
@@ -0,0 +1,76 @@
...
+            logger.error(error_msg, function_name, op, extra=dict(data=data, response=response))
+            raise
+
+        lambda_result: str = cast(str, response.get("result"))
+        if lambda_result in LambdaResult.successes():
+            return response
+        if lambda_result in allowed_error_results:
+            return response
+        if raise_if_error_resp:
+            raise LambdaError(function_name, op, lambda_result, response)  # type: ignore
```
redlickigrzegorz, line 71: [#981451895](https://github.com/pdyba/lambdalizator/pull/64#discussion_r981451895):
> You should resolve the problem instead of disabling type checking (rel. [#965942904](https://github.com/pdyba/lambdalizator/pull/64#discussion_r965942904)) 😉

lbz/lambdas/client.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # f-string used directly to keep messages unique
```
redlickigrzegorz, line 73: [#981463529](https://github.com/pdyba/lambdalizator/pull/64#discussion_r981463529):
```suggestion
        # f-string used directly to keep messages unique from the monitoring/tracking perspective
```

lbz/lambdas/broker.py
```diff
@@ -0,0 +1,42 @@
...
+        self.event = event
+        self.context = context
+        self.mapper = mapper
+        self.data = event.get("data")
+
+    def handle(self) -> LambdaResponse:
+        handler = self._get_handler()
+        if isinstance(handler, dict):
+            return handler
+        if self.data:
```
redlickigrzegorz, line 27: [#966041983](https://github.com/pdyba/lambdalizator/pull/64#discussion_r966041983):
> This is not correct as it can be e.g. an empty dict - as for me, you should set the whole `event` at the object level and try to fetch data here directly.

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,40 @@
...
+        func_1 = MagicMock(return_value="some_data")
+        mapper = {"x": func_1}
+        event = {"op": "x", "data": {"y": 1}}
+
+        resp = LambdaBroker(mapper, event).react()  # type: ignore
+
+        assert resp == "some_data"
+        func_1.assert_called_once_with({"y": 1})
+
+    def test_broker_responds_raises_implemented_when_event_type_is_not_recognized(self) -> None:
```
redlickigrzegorz, line 19: [#981509875](https://github.com/pdyba/lambdalizator/pull/64#discussion_r981509875):
```suggestion
    def test_broker_raises_error_when_event_type_is_not_recognized(self) -> None:
```

lbz/lambdas/__init__.py
pdyba: Sue but that will be sorted when merging

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,65 @@
...
+
+        func_1.assert_not_called()
+        assert resp == {
+            "result": LambdaResult.BAD_REQUEST,
+            "message": "Lambda execution error: Missing 'op' field in the event.",
+        }
+
+    def test_broker_handles_expected_error_outcome(self) -> None:
+        def func_1(_event: dict) -> None:
+            raise NotFound()
```
redlickigrzegorz, line 46: Provide the `error_code` here to confirm that it is being forwarded.

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,65 @@
...
+
+        mapper = {"x": func_1}
+        event = {"op": "x", "data": {"y": 1}}
+        resp = LambdaBroker(mapper, event).react()  # type: ignore
```
redlickigrzegorz, line 50: Here, `mypy` needs to be satisfied instead of ignored.

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,65 @@
...
+
+        assert resp == {"result": LambdaResult.SERVER_ERROR, "message": NotFound.message}
+
+    def test_broker_handles_unexpected_error_outcome(self) -> None:
+        def func_1(_event: dict) -> None:
+            raise EOFError("oops")
+
+        mapper = {"x": func_1}
+        event = {"op": "x", "data": {"y": 1}}
+        resp = LambdaBroker(mapper, event).react()  # type: ignore
```
redlickigrzegorz, line 60: In general, you should enable `mypy` for this new test package 😉 
```Makefile
mypy lbz tests/test_events tests/test_lambdas setup.py
```

lbz/lambdas/broker.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def handle(self) -> LambdaResponse:
+        if not (op := self.event.get("op")):
+            return lambda_error_response(
+                result=LambdaResult.BAD_REQUEST,
+                error_message="Lambda execution error: Missing 'op' field in the event.",
+            )
+        try:
+            handler = self._get_handler(op)
+            if data := self.event.get("data"):
```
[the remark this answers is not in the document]
redlickigrzegorz: If we check the presence of the key, we will have to adjust `LambdaClient` to stop sending it when not needed.

That is the decision from the past (from the last year, I guess) that `None` means no data 🤷

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+from lbz.lambdas import LambdaBroker, LambdaResponse, LambdaResult, lambda_ok_response
+
+
+def simple_func(*_event: dict) -> LambdaResponse:
+    return lambda_ok_response()
+
+
+class TestEventBroker:
+    def test_broker_works_properly_when_data_is_provided(self) -> None:
+        def func(_event: dict) -> LambdaResponse:
```
redlickigrzegorz, line 11: Also, this should not be named `_event` as handlers accept only `data` (part of the event - payload).

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {
+            "result": LambdaResult.OK,
+            "data": {"some": "data"},
+        }
+
+    def test_broker_works_properly_when_no_data_is_provided(self) -> None:
+        mapper = {"x": simple_func}
+        event = {"op": "x"}
```
redlickigrzegorz, line 26: This is not correct, `LambdaClient` always sends `data` but it can be equal to `None` 😉

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {
+            "result": LambdaResult.OK,
+            "data": {"some": "data"},
+        }
+
+    def test_broker_works_properly_when_no_data_is_provided(self) -> None:
+        mapper = {"x": simple_func}
```
redlickigrzegorz, line 25: Also, you need to create a handler that does not accept `data` at all - `simple_func` doesn't care.

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+        event = {"op": "x"}
+
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {
+            "result": LambdaResult.OK,
+        }
+
+    def test_broker_responds_with_bad_request_event_type_is_not_recognized(self) -> None:
```
redlickigrzegorz, line 34:
```suggestion
    def test_broker_responds_with_bad_request_when_op_is_not_recognized(self) -> None:
```

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+        event = {"op": "y", "data": {"z": 1}}
+
+        response = LambdaBroker(mapper, event).react()
+
+        assert response == {
+            "result": LambdaResult.BAD_REQUEST,
+            "message": "Handler for y was no implemented",
+        }
+
+    def test_broker_responds_no_op_key(self) -> None:
```
redlickigrzegorz, line 45:
```suggestion
    def test_broker_responds_with_bad_request_when_no_op_key(self) -> None:
```

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+        event = {"data": {"y": 1}}
+
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {
+            "result": LambdaResult.BAD_REQUEST,
+            "message": "Missing 'op' field in the event.",
+        }
+
+    def test_broker_handles_unexpected_lbfw_error_outcome(self) -> None:
```
redlickigrzegorz, line 56: `lbfw`? Did you mean `lbz_exception`? 🤔

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {"result": LambdaResult.SERVER_ERROR, "message": NotFound.message}
+
+    def test_broker_handles_unexpected_lbfw_error_outcome_with_error_code(self) -> None:
+        class MyNotFound(NotFound):
+            error_code = "NN404"
+
+        def func(_event: dict) -> LambdaResponse:
+            raise MyNotFound()
```
redlickigrzegorz, line 71:
```suggestion
        def func(_data: dict) -> LambdaResponse:
            raise NotFound(error_code="NN404")
```

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,94 @@
...
+        event = {"op": "x", "data": {"y": 1}}
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {
+            "result": LambdaResult.SERVER_ERROR,
+            "message": NotFound.message,
+            "error_code": "NN404",
+        }
+
+    def test_broker_handles_unexpected_error_outcome(self) -> None:
```
redlickigrzegorz, line 83: You should also confirm that these unexpected Exceptions are logged.

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,141 @@
...
+        assert resp == {"result": LambdaResult.SERVER_ERROR, "message": NotFound.message}
+        assert caplog.record_tuples == [
+            (
+                "lbz.lambdas.broker",
+                logging.ERROR,
+                'Unexpected error in "func" function!',
+            )
+        ]
+
+    def test_broker_handles_unexpected_lbfw_error_outcome_with_error_code(
```
redlickigrzegorz, line 100: `lbz_exception` exactly as above.

lbz/lambdas/broker.py
```diff
@@ -0,0 +1,40 @@
...
+        context: object = None,
+    ) -> None:
+        self.event = event
+        self.context = context
+        self.mapper = mapper
+
+    def handle(self) -> LambdaResponse:
+        if not (op := self.event.get("op")):
+            logger.error('Missing "op" field in the processed event: %r', self.event)
+            return lambda_error_response(LambdaResult.BAD_REQUEST, 'Missing "op" field in event.')
```
redlickigrzegorz, line 26: As the final word, I think this should be `CONTRACT_ERROR` as this will be used for internal communication. Here, we are both client and server, so, when something like this failed, it means we did not meet the contract between services (lambdas):
```suggestion
            return lambda_error_response(LambdaResult.CONTRACT_ERROR, 'Missing "op" field.')
```

lbz/lambdas/broker.py
```diff
@@ -0,0 +1,40 @@
...
+        if not (handler := self.mapper.get(op)):
+            logger.error('No handler declared for requested operation: "%s"', op)
+            return lambda_error_response(LambdaResult.BAD_REQUEST, f'"{op}" is not implemented.')
```
redlickigrzegorz, line 29: Same here, but if we are careful we will never see this error 😉 
```suggestion
            return lambda_error_response(LambdaResult.CONTRACT_ERROR, f'"{op}" not implemented.')
```

tests/test_lambdas/test_broker.py
```diff
@@ -0,0 +1,141 @@
...
+        mapper = {"x": no_data_func}
+        event = {"op": "x", "data": None}
+
+        resp = LambdaBroker(mapper, event).react()
+
+        assert resp == {
+            "result": LambdaResult.OK,
+        }
+
+    def test_broker_responds_with_bad_request_when_op_is_not_recognized(
```
redlickigrzegorz, line 41: Don't forget to update the name of the tests after the last changes.

redlickigrzegorz (approved):

PR #69, opened 2022-09-28
adds config management

pdyba:


lbz/configuration/aws_ssm.py
```diff
@@ -0,0 +1,9 @@
+from typing import Any
+
+from lbz.aws_boto3 import client
+
+
+class SSM:
```
redlickigrzegorz, line 6: [#961433214](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961433214):
> Why is `aws_ssm` placed in the configuration? Personally, I am not sure if I would look for that here 🤔
> 
> I think we can put it at the top level, next to `aws_boto3.py` 😉

lbz/configuration/aws_ssm.py
```diff
@@ -0,0 +1,9 @@
...
+    @staticmethod
+    def get_parameter(name: str) -> Any:
```
redlickigrzegorz, line 8: [#961439374](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961439374):
> Isn't it true that the return value is always string? 🤔

tests/test_configuration/test_aws_ssm.py
```diff
@@ -0,0 +1,14 @@
+from unittest.mock import MagicMock, patch
+
+from lbz.aws_boto3 import client
+from lbz.configuration import SSM
+
+
+@patch.object(client.ssm, "get_parameter")
+def test_get_parameter(mocked_get_parameter: MagicMock) -> None:
```
redlickigrzegorz, line 8: [#961442723](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961442723):
```suggestion
def test__get_parameter__returns_value_fetched_from_aws(mocked_get_parameter: MagicMock) -> None:
```

tests/test_configuration/test_aws_ssm.py
```diff
@@ -0,0 +1,14 @@
+from unittest.mock import MagicMock, patch
+
+from lbz.aws_boto3 import client
+from lbz.configuration import SSM
+
+
+@patch.object(client.ssm, "get_parameter")
```
redlickigrzegorz, line 7: [#961444754:](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961444754)
> I would mock here the whole ssm, as it is a little bit confusing that you mock get_parameter during testing get_parameter 😉

tests/test_configuration/test_aws_ssm.py
```diff
@@ -0,0 +1,14 @@
...
+def test_get_parameter(mocked_get_parameter: MagicMock) -> None:
+    mocked_get_parameter.return_value = {"Parameter": {"Value": "x"}}
+
+    value = SSM.get_parameter("path/to/param")
```
redlickigrzegorz, line 11: [#961445636](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961445636):
> This is rather a name, not a path 😉

lbz/configuration/configs.py
```diff
@@ -0,0 +1,58 @@
...
+            else:
+                self._value = self.parser(val) if self.parser else val
+        return self._value
+
+    @property
+    def value(self) -> Any:
+        return self.get_value()
+
+
+class KeyValue(BaseConfig):
```
redlickigrzegorz, line 39: [#961457397](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961457397):
> This can be simply replaced by a globally declared variable - at least I would do that this way to simplify declaring/usage 😉

tests/__init__.py
```diff
@@ -9,8 +9,11 @@
...
 )
 
+# JWT related test config
 os.environ["AUTH_REMOVE_PREFIXES"] = "1"
 os.environ["ALLOWED_PUBLIC_KEYS"] = json.dumps({"keys": [SAMPLE_PUBLIC_KEY]})
 os.environ["ALLOWED_AUDIENCES"] = ",".join([str(uuid4()), str(uuid4())])
 os.environ["ALLOWED_ISS"] = "test-issuer"
+# AWS related test config
 os.environ["AWS_LAMBDA_FUNCTION_NAME"] = "million-dollar-lambda"
+os.environ["AWS_DEFAULT_REGION"] = "us-west-2"
```
redlickigrzegorz, line 19: [#961430505](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961430505):
> It would be better to set these all env variables with the usage of the fixture.
> 
> But you can place TODO to do it in the future if you wish 😉

tests/conftest.py
```diff
@@ -66,12 +66,12 @@ def sample_event() -> APIGatewayEvent:
 
 
 @pytest.fixture(scope="session", name="jwt_partial_payload")
-def jwt_partial_payload_fixture() -> dict:
+def jwt_partial_payload_fixture(allowed_audiences: list) -> dict:
```
redlickigrzegorz, line 69:
```suggestion
def jwt_partial_payload_fixture(allowed_audiences: List[str]) -> dict:
```

lbz/configuration/configs.py
```diff
@@ -0,0 +1,58 @@
...
+            val = val if val is not None else self.default
+            if val is None:
+                self._value = None
+            else:
+                self._value = self.parser(val) if self.parser else val
+        return self._value
+
+    @property
+    def value(self) -> Any:
+        return self.get_value()
```
redlickigrzegorz, line 36: I don't see many benefits from having three methods if here, only `.get_value()` is triggered. It could be implemented this way:
```python
@property
def value(self) -> Any:
    if self._value is None:
        val = self.get_value()
        ...

@abstractmethod
def get_value(self) -> Any:
    pass
```

lbz/configuration/configs.py
```diff
@@ -0,0 +1,58 @@
+from abc import abstractmethod
+from os import getenv
+from typing import Any, Callable, Optional
+
+from lbz.configuration.aws_ssm import SSM
+
+
+class BaseConfig:
```
redlickigrzegorz, line 8: The name should be updated, exactly as you did below.

lbz/configuration/configs.py
```diff
@@ -0,0 +1,58 @@
...
+    ):
+        super().__init__("key", parser=None, default=None)
+        self._value = value
+
+    def getter(self) -> Any:
+        return self._value
+
+
+class EnvValue(BaseConfig):
+    def getter(self) -> Any:
```
redlickigrzegorz, line 52: Whenever we declare a new configuration variable, the type needs to be strong, we cannot rely on `Any` or `Optional` because it is problematic in further code that uses this variable 😕

redlickigrzegorz (changes requested):

redlickigrzegorz:
I mentioned also two things about requirements in the already closed pull request #63 that I would consider doing here 😉 

[#961427269](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961427269):
> I think we should remove the version restriction here: [requirements-dev.in#L4](https://github.com/pdyba/lambdalizator/blob/7e8cd28742668c43a24f96d4be8549f90c8c4ea3/requirements-dev.in#L4)

[#961438691](https://github.com/pdyba/lambdalizator/pull/63#discussion_r961438691):
> Now, I also see that we have too many things put here - we don't use `dynamodb`, `s3`, `sns`, and `sqs` in Lambdalizator, do we?

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+
+from lbz.aws_boto3 import SSM
+
+T = TypeVar("T")
+
+
+class BaseValue(Generic[T]):
+    def __init__(
+        self,
+        key: str,
```
redlickigrzegorz, line 13: Maybe `name`? Key is strongly related to dictionaries and env variables 🤔 

But it is only a suggestion.

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        parser: Optional[Callable[..., T]] = None,
```
redlickigrzegorz, line 14: I think every parser should be exactly the same, accepting string as a single positional argument 🤔

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        self.default = default
+        self._value: T
+
+    @abstractmethod
+    def getter(self) -> Optional[T]:
+        pass
+
+    @property
+    def value(self) -> T:
+        if not hasattr(self, "_value"):
```
redlickigrzegorz, line 28: I think I prefer to set `_value` equal to `None` in init instead of relying on this line and have the problems below 😉

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+class BaseValue(Generic[T]):
+    def __init__(
+        self,
+        key: str,
+        parser: Optional[Callable[..., T]] = None,
+        default: T = None,
+    ):
+        self.key = key
+        self.parser = parser
+        self.default = default
```
redlickigrzegorz, line 19: I would make all these attributes protected - we don't want to change them from outside anyway:
```suggestion
        self._key = key
        self._parser = parser
        self._default = default
```

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        self._value: T
+
+    @abstractmethod
+    def getter(self) -> Optional[T]:
```
redlickigrzegorz, line 23: I think `getter` always gets value in a raw version: `Optional[str]` 😉

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        pass
+
+    @property
+    def value(self) -> T:
+        if not hasattr(self, "_value"):
+            val = self.getter()
+            val = val if val is not None else self.default
+            if val is None:
+                self._value = None  # type:ignore
```
redlickigrzegorz, line 32: `mypy` is right, you said above that you don't return `None` from this method.

redlickigrzegorz, line 32: And in general, we cannot return `None` because it puts us exactly in the same place as before - I mean that the rest of our code won't be able to rely on types coming from configuration 😐

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+    @abstractmethod
+    def getter(self) -> Optional[T]:
+        pass
+
+    @property
+    def value(self) -> T:
+        if not hasattr(self, "_value"):
+            val = self.getter()
+            val = val if val is not None else self.default
+            if val is None:
```
redlickigrzegorz, line 31: Also, I think you need to return to the idea of raising an Exception when the config value is missing.

Otherwise, the code will fail unexpectedly in places where `None` is a bad value to run with.

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+                self._value = None  # type:ignore
+            else:
+                self._value = self.parser(val) if self.parser else val
```
redlickigrzegorz, line 34: You should consider setting a default parser, this should reduce if conditions in this method 😉

lbz/configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        return self._value
+
+
+class EnvValue(BaseValue):
+    def getter(self) -> Optional[Any]:
```
redlickigrzegorz, line 39: We cannot have anything else than a string in env variables, can we?
```suggestion
    def getter(self) -> Optional[str]:
```

lbz/configuration.py
redlickigrzegorz, line 34: What do you think about raising a dedicated error when parsing fails? 🤔 

This could help us see the problem more directly instead of `TypeError` or `ValueError`.

tests/test_configuration.py
```diff
@@ -0,0 +1,81 @@
...
+from unittest.mock import MagicMock, patch
+
+from lbz.aws_boto3 import SSM
+from lbz.configuration import BaseValue, EnvValue, SSMValue
+
+
+class TestBaseConfig:
+    @patch.object(BaseValue, "getter", autospec=True)
+    def test_if_getter_was_used(self, mocked_getter: MagicMock) -> None:
+        cfg = BaseValue("key")  # type: ignore
```
redlickigrzegorz, line 12: I didn't check the tests yet but definitely, you cannot do things like this.

`BaseValue` is an abstract class so if you want to rely on this you need to create a new one 😉

tests/test_configuration.py
```diff
@@ -0,0 +1,81 @@
...
+
+        cfg.value  # pylint: disable=pointless-statement
```
redlickigrzegorz, line 14: The same here, this should return real value which you can check.

tests/test_configuration.py
```diff
@@ -0,0 +1,81 @@
...
+
+        mocked_getter.assert_called_once()
+
+    def test_returns_none_when_no_value_set(self) -> None:
+        cfg = BaseValue("test")  # type: ignore
```
redlickigrzegorz, line 19: Also, enable `mypy` for these new tests.

redlickigrzegorz (changes requested):

lbz/configuration.py
pdyba: that the point its the key to get the value

lbz/configuration.py
pdyba: That's more tricky actually...

lbz/configuration.py
pdyba: Then return Type becomes optional and we land were we started

lbz/configuration.py
```diff
@@ -0,0 +1,60 @@
...
+class ConfigValue(Generic[T], metaclass=ABCMeta):
+    """Mandatory Configuration
+
+    This class is not supporting None as outcome value.
+    """
+
+    def __init__(
+        self,
+        key: str,
+        parser: Callable[[str], T] = str,  # type: ignore
```
redlickigrzegorz, line 20: Once we changed `getter()` to return `Any`, we need to adjust the parser as well:
```suggestion
        parser: Callable[[Any], T] = str,  # type: ignore
```

lbz/configuration.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @abstractmethod
+    def getter(self) -> Any:
+        pass
+
+    @property
+    def value(self) -> T:
+        if self._value is None:
+            val = self.getter()
+            if val is not None:
```
redlickigrzegorz, line 36: New `pylint` configuration will point that out anyway:
```suggestion
            if (val := self.getter()) is not None:
```

lbz/configuration.py
```diff
@@ -0,0 +1,60 @@
...
+            else:
+                raise ConfigurationMissingError(f"'{self._key}' was not defined.")
+
+        return self._value
+
+    def _parse_value(self, value: Any) -> T:
+        try:
+            return self._parser(value)
+        except Exception as error:
+            message = f"'{self._key}' could not be parsed with '{self._parser}'"
```
redlickigrzegorz, line 49: This will be moved directly to the Exception but I think much more valuable is to put here the value that we tried to parse, instead of the representation of the parser function 😉

lbz/configuration.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class EnvValue(ConfigValue):
+    def getter(self) -> Optional[str]:
+        return getenv(self._key)
+
+
+class SSMValue(ConfigValue):
+    def getter(self) -> Optional[str]:
+        return SSM.get_parameter(self._key)
```
redlickigrzegorz, line 60: `SSM.get_parameter` is not compatible with `ConfigValue` as it always tries to return `str`.

If the value is not set, we will get an error from AWS or `KeyError` 🤔

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+    def getter(self) -> str:
+        return "test-value"
+
+
+class MyNoneConfigValue(ConfigValue):
+    def getter(self) -> None:
+        return None
+
+
+class TestBaseConfig:
```
redlickigrzegorz, line 22:
```suggestion
class TestConfigValue:
```

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_if_getter_was_used(
+        self,
+    ) -> None:
```
redlickigrzegorz, line 25:
```suggestion
    def test_returns_value_fetched_by_getter(self) -> None:
```

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+        cfg = MyConfigValue("key")
+
+        assert cfg.value == "test-value"
+
+    def test_returns_default_when_no_value_set(self) -> None:
+        cfg = MyNoneConfigValue("test", default=42, parser=int)
+
+        assert cfg.value == 42
+
+    def test_raises_missing_congiration_when_value_is_none(self) -> None:
```
redlickigrzegorz, line 35: Typo:
```suggestion
    def test_raises_missing_configuration_when_value_is_none(self) -> None:
```

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+
+        with pytest.raises(ConfigurationMissingError, match="'test' was not defined."):
+            cfg.value  # pylint: disable=pointless-statement
+
+    def test_if_parser_was_used(self) -> None:
+        class MyIntConfigValue(ConfigValue):
+            def getter(self) -> int:
+                return 1
+
+        cfg = MyIntConfigValue("key")
```
redlickigrzegorz, line 46: I would rather use a custom parser here in this test to show more directly what you expect.

Now, you rely on a default behavior that is tested multiple times in other places.

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+
+        assert cfg.value == 42
+
+    def test_raises_missing_congiration_when_value_is_none(self) -> None:
+        cfg = MyNoneConfigValue("test")
+
+        with pytest.raises(ConfigurationMissingError, match="'test' was not defined."):
+            cfg.value  # pylint: disable=pointless-statement
+
+    def test_if_parser_was_used(self) -> None:
```
redlickigrzegorz, line 41:
```suggestion
    def test_modifies_config_value_using_declared_parser_function(self) -> None:
```

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @patch.object(MyConfigValue, "getter", autospec=True)
+    def test_if_getter_was_used_only_once(self, mocked_getter: MagicMock) -> None:
+        mocked_getter.return_value = "42"
+        cfg = MyConfigValue("key")
+
+        assert cfg.value == "42"
+        assert cfg.value == "42"
+
+        mocked_getter.assert_called_once()
```
redlickigrzegorz, line 58: You don't have to completely mock the getter to check it was used only once 😉 
```suggestion
    def test_fetches_original_value_only_once(self) -> None:
        cfg = MyConfigValue("key")

        with patch.object(MyConfigValue, "getter", wraps=cfg.getter) as mocked_getter:
            assert cfg.value == "test-value"
            assert cfg.value == "test-value"

        mocked_getter.assert_called_once()
```

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_raises_if_parser_failed(self) -> None:
+        class MyFailingConfigValue(ConfigValue):
+            def getter(self) -> tuple:
+                return 1, 2
+
+        cfg = MyFailingConfigValue("key", parser=int)
+
+        with pytest.raises(
+            ConfigurationParsingError,
+            match="'key' could not be parsed with '<class 'int'>'",
```
redlickigrzegorz, line 69: Please, don't create multiline context managers, you can always declare a variable for the expected message.

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_gets_value_from_env_with_list_parser(self) -> None:
+        env_cfg = EnvValue("RANDOM_VAR", parser=list)
+
+        assert env_cfg.value == ["1", "2"]
+
+    @patch.dict(environ, {"RANDOM_VAR": '{"test": "ttt"}'}, clear=True)
+    def test_gets_value_from_env_with_json_parser(self) -> None:
+        env_cfg = EnvValue("RANDOM_VAR", parser=json.loads)
+
+        assert env_cfg.value == {"test": "ttt"}
```
redlickigrzegorz, line 97: These three tests are testing the behavior of the base class, not `EnvConfig`.

They are ok to keep but should be moved to the above class and adjusted to use `ConfigValue` instead 😉

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+
+        with pytest.raises(
+            ConfigurationParsingError,
+            match="'key' could not be parsed with '<class 'int'>'",
+        ):
+            cfg.value  # pylint: disable=pointless-statement
+
+
+class TestEnvConfig:
+    @patch.dict(environ, {"RANDOM_VAR": "12.2.1"}, clear=True)
```
redlickigrzegorz, line 75: Is there any significant benefit from clearing all environment variables before setting this one?

tests/test_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+    @patch.dict(environ, {"RANDOM_VAR": '{"test": "ttt"}'}, clear=True)
+    def test_gets_value_from_env_with_json_parser(self) -> None:
+        env_cfg = EnvValue("RANDOM_VAR", parser=json.loads)
+
+        assert env_cfg.value == {"test": "ttt"}
+
+
+class TestSSMConfig:
+    @patch.object(SSM, "get_parameter")
+    def test_getter_calls_ssn_with_specific_key(self, mocked_get_parameter: MagicMock) -> None:
```
redlickigrzegorz, line 102: Typo:
```suggestion
    def test_getter_calls_ssm_with_specific_key(self, mocked_get_parameter: MagicMock) -> None:
```

lbz/configuration.py
redlickigrzegorz: On the other hand, we have `SSMValue` which operates with the name of the parameter 😉 

But I won't be complaining about it - the key is ok enough for now 👍

lbz/configuration.py
redlickigrzegorz: As it turned out, it wasn't that tricky but in the newest version, it should accept `Any` 🙂

lbz/configuration.py
redlickigrzegorz: Fortunately, the newest version handled that correctly 👍

tests/test_configuration.py
redlickigrzegorz: Or, you can convert all the tests to use `EnvConfig` instead - not an ideal idea but we would get rid of all custom classes created just for tests 🤔 

The final decision is on you but personally, I would consider doing it, just to stay consistent.

tests/test_configuration.py
pdyba: Just want to be sure that's the only one.

tests/test_configuration.py
```diff
@@ -0,0 +1,93 @@
...
+class TestEnvConfig:
+    @patch.dict(environ, {"RANDOM_VAR": "12.2.1"}, clear=True)
+    def test_gets_value_from_env(self) -> None:
+        env_cfg = EnvValue("RANDOM_VAR")
+
+        assert env_cfg.value == "12.2.1"
+
+
+class TestSSMConfig:
+    @patch.object(SSM, "get_parameter")
```
redlickigrzegorz, line 86:
```suggestion
    @patch.object(SSM, "get_parameter", autospec=True)
```

redlickigrzegorz (approved):

tests/test_configuration.py
```diff
@@ -0,0 +1,87 @@
...
+        with pytest.raises(MissingConfigValue, match="'test' was not defined."):
+            cfg.value  # pylint: disable=pointless-statement
+
+    @patch.object(EnvValue, "getter", MagicMock(return_value=1))
+    def test_modifies_config_value_using_declared_parser_function(self) -> None:
+        cfg = EnvValue("RANDOM_VAR", lambda a: a * 10)
+
+        assert cfg.value == 10
+
+    @patch.object(EnvValue, "getter", MagicMock(return_value=(1, 2)))
```
redlickigrzegorz, line 43: You should not mock getter, but rather adjust the test to show expected behavior based on the string value 😉

PR #70, opened 2022-10-06
Importing of configuration fix, updates readme

pdyba:


lbz/_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+from lbz.configuration import EnvValue
+
+# LBZ configuration
+LBZ_DEBUG_MODE = EnvValue("LBZ_DEBUG_MODE", default=False, parser=lambda a: a == "true")
+LOGGING_LEVEL = EnvValue("LOGGING_LEVEL", default="INFO")
+CORS_HEADERS = EnvValue("CORS_HEADERS", default=[], parser=lambda a: a.split(","))
+CORS_ORIGIN = EnvValue("CORS_ORIGIN", default=[], parser=lambda a: a.split(","))
+
+# AWS related configuration
+AWS_LAMBDA_FUNCTION_NAME = EnvValue("AWS_LAMBDA_FUNCTION_NAME", default="lbz-event-api").value
```
redlickigrzegorz, line 12: Once this variable became more generic, I don't think we should put the default value to `lbz-event-api`.

Maybe, we shouldn't use the default value at all - this would make it required to be set if this part of the code is used 🤔

lbz/_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+EVENTS_BUS_NAME = EnvValue(
+    "EVENTS_BUS_NAME", default=f"{AWS_LAMBDA_FUNCTION_NAME}-event-bus"
```
redlickigrzegorz, line 14: The same here - I don't think the default value is good here.

lbz/_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+).value
```
redlickigrzegorz, line 15: Also, resigning from lazy-loading may produce unexpected problems in tests (depending on when exactly these values are set) 🤔

lbz/_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Authorization configuration
+PUBLIC_KEYS = EnvValue("ALLOWED_PUBLIC_KEYS", parser=lambda a: json.loads(a)["keys"])
+ALLOWED_AUDIENCES = EnvValue("ALLOWED_AUDIENCES", parser=lambda a: a.split(","))
+ALLOWED_ISS = EnvValue("ALLOWED_ISS", default="")
+REMOVE_PREFIXES = EnvValue("AUTH_REMOVE_PREFIXES", default=False, parser=bool)
```
redlickigrzegorz, line 21: I'm not sure if to use `bool` as a parser - it means that only an empty string or an unset value will give us `False`.

On the other hand e.g. `0` and` false` will stand for `True` 🤔

redlickigrzegorz, line 21: Also, shouldn't it be named `AUTH_REMOVE_PREFIXES` - the same as the env value?

lbz/authentication.py
```diff
@@ -1,14 +1,11 @@
 """
 JWT based Authentication module.
 """
-import os
-
+from lbz._cfg import REMOVE_PREFIXES
```
redlickigrzegorz, line 4: Previously, it was a boolean value but now here, you get the `EnvValue` object - you need to use its `.value` to work with it correctly below.

lbz/dev/server.py
```diff
@@ -6,16 +6,12 @@
...
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
-from os import environ
 from threading import Thread
 from typing import Tuple, Type, Union
 
 from lbz.dev.misc import APIGatewayEvent
 from lbz.resource import Resource
 
-if environ.get("LBZ_DEBUG_MODE") is None:
-    environ["LBZ_DEBUG_MODE"] = "true"
```
redlickigrzegorz, line 17: It makes a little sense to have it in the dev server 🤔 

At least I remember you added it to always enable debug mode here.

lbz/misc.py
```diff
@@ -131,7 +130,7 @@ def deep_update(dict_to_update: dict, update_data: dict) -> None:
 
 
 def is_in_debug_mode() -> bool:
-    return environ.get("LBZ_DEBUG_MODE", "").lower() == "true"
+    return cast(bool, LBZ_DEBUG_MODE.value)
```
redlickigrzegorz, line 133: If we need to cast the type of `.value`, this means our `ConfigValue` is wrongly implemented 😞

redlickigrzegorz: Or the parser is not specific enough 🤔

lbz/_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+CORS_ORIGIN = EnvValue("CORS_ORIGIN", default=[], parser=lambda a: a.split(","))
+
+# AWS related configuration
+AWS_LAMBDA_FUNCTION_NAME = EnvValue("AWS_LAMBDA_FUNCTION_NAME", default="lbz-event-api").value
+EVENTS_BUS_NAME = EnvValue(
+    "EVENTS_BUS_NAME", default=f"{AWS_LAMBDA_FUNCTION_NAME}-event-bus"
+).value
+
+# Authorization configuration
+PUBLIC_KEYS = EnvValue("ALLOWED_PUBLIC_KEYS", parser=lambda a: json.loads(a)["keys"])
```
redlickigrzegorz, line 18: The same here, shouldn't it be named `ALLOWED_PUBLIC_KEYS`?

lbz/resource.py
```diff
@@ -89,7 +90,11 @@ def __repr__(self) -> str:
         return f"<Resource {self.method} @ {self.urn} >"
 
     def _load_configuration(self) -> None:
-        self.auth_enabled = env.get("ALLOWED_PUBLIC_KEYS") or env.get("ALLOWED_AUDIENCES")
```
redlickigrzegorz, line 92: Previously, this variable was set based on both `ALLOWED_PUBLIC_KEYS` and `ALLOWED_AUDIENCES`.

lbz/resource.py
```diff
@@ -89,7 +90,11 @@ def __repr__(self) -> str:
...
+        try:
+            PUBLIC_KEYS.value  # pylint: disable=pointless-statement
+            self.auth_enabled = True
+        except MissingConfigValue:
```
redlickigrzegorz, line 96: We should never base on this Exception, so, is `ALLOWED_PUBLIC_KEYS` required or not?

lbz/resource.py
```diff
@@ -154,13 +159,10 @@ def __init__(
     ):
         # TODO: adjust the rest of the arguments in the near future too.
         super().__init__(event)
-        cors_headers = cors_headers or []
-        if not cors_headers and (env_headers := env.get("CORS_HEADERS")):
-            cors_headers = env_headers.split(",")
+        if not cors_headers:
+            cors_headers = cast(list, CORS_HEADERS.value)
```
redlickigrzegorz, line 163: Here we have the same problem - it is very possible that we should not use lambda functions as parsers 🤔

lbz/resource.py
```diff
@@ -154,13 +159,10 @@ def __init__(
     ):
         # TODO: adjust the rest of the arguments in the near future too.
         super().__init__(event)
-        cors_headers = cors_headers or []
```
redlickigrzegorz, line 157: This line was pretty important as the code below assumes that `cors_headers` is always a list.

tests/test_authentication.py
```diff
@@ -29,14 +30,16 @@ def test_decoding_user_raises_unauthorized_when_invalid_token(user_token: str) -
 
 
 def test_decoding_user_raises_unauthorized_when_invalid_audience(user_token: str) -> None:
-    with pytest.raises(Unauthorized), patch("lbz.jwt_utils.ALLOWED_AUDIENCES", [str(uuid4())]):
+    with pytest.raises(Unauthorized), patch(
+        "lbz.jwt_utils.ALLOWED_AUDIENCES", MockedConfig(str(uuid4()), parser=lambda a: [a])
```
redlickigrzegorz, line 34: It was done this way before as this variable was declared globally. Now, when we base it on configuration values, it shouldn't be needed.

But to remove all these mocks or to make them better, I would need to investigate what they really want to replace 🤔

tests/test_jwt_utils.py
```diff
@@ -63,12 +63,6 @@ def test_expired_jwt(self) -> None:
         with pytest.raises(Unauthorized, match="Your token has expired. Please refresh it."):
             decode_jwt(jwt_token)
 
-    @patch("lbz.jwt_utils.get_matching_jwk", MagicMock(return_value={}))
-    @patch("lbz.jwt_utils.ALLOWED_AUDIENCES", [])
-    def test_empty_allowed_audiences(self) -> None:
-        with pytest.raises(RuntimeError, match="ALLOWED_AUDIENCES"):
```
redlickigrzegorz, line 69: In general, the logic remains but we raise a different Exception 🤔 

I would leave these tests and only adjust them, otherwise, we can easily misconfigure lbz by setting a default value for these two variables.

tests/test_resource.py
```diff
@@ -370,7 +371,10 @@ def test_cors_headers_param_request(self) -> None:
             ALLOW_ORIGIN_HEADER: ORIGIN_EXAMPLE,
         }
 
-    @patch.dict(environ, {"CORS_HEADERS": "X-PRN-KEY,X-PRN-TOKEN"})
+    @patch(
+        "lbz.resource.CORS_HEADERS",
+        MockedConfig("x", parser=lambda _a: ["X-PRN-KEY", "X-PRN-TOKEN"]),
```
redlickigrzegorz, line 376: I have a feeling that we need to add the possibility to clear configuration 🤔

Each test should be treated more or less separately, otherwise, workarounds like this one will be added randomly in the code - unfortunately, this is the wrong approach.

tests/conftest.py
```diff
@@ -20,6 +21,11 @@
 from tests.utils import encode_token
 
 
+class MockedConfig(ConfigValue):
```
redlickigrzegorz, line 24: Based on the above comments, this class should be removed.

redlickigrzegorz (changes requested):

lbz/resource.py
pdyba: I know but it was wrong.

lbz/_cfg.py
pdyba: Agreed remove default

lbz/resource.py
pdyba: and it is - default ensures that.

lbz/configuration.py
```diff
@@ -1,13 +1,30 @@
...
 T = TypeVar("T")
 
 
+class ConfigParser:
+    @staticmethod
+    def split_by_comma(cfg: str) -> List[str]:
+        return cfg.split(",")
+
+    @staticmethod
+    def env_bool(cfg: str) -> bool:
```
redlickigrzegorz, line 18: Why `env`? Exactly the same string could be stored in SSM 😉

redlickigrzegorz: Also, to keep naming consistent, it should be rather named as a command e.g. `cast_to_bool` or something similar.

lbz/configuration.py
```diff
@@ -1,13 +1,30 @@
...
+        if cfg.lower() in ("true", "1"):
+            return True
+        return False
+
+    @staticmethod
+    def load_jwt_keys(cfg: str) -> List[dict]:
+        return cast(list, json.loads(cfg)["keys"])
```
redlickigrzegorz, line 25: Using `cast` is not a silver bullet. First you should think why `mypy` cannot recognize the type and try to help him 😉 

e.g.
```python
deserialized_value: Dict[str, List[dict]] = json.loads(value)
return deserialized_value["keys"]
```

You can even create a custom `TypedDict` instead of declaring it this way - it might be useful in other places.

lbz/configuration.py
```diff
@@ -1,13 +1,30 @@
...
+        return cfg.split(",")
+
+    @staticmethod
+    def env_bool(cfg: str) -> bool:
+        if cfg.lower() in ("true", "1"):
+            return True
+        return False
+
+    @staticmethod
+    def load_jwt_keys(cfg: str) -> List[dict]:
```
redlickigrzegorz, line 24: Also, it would be good to write simple tests for these parsers - it will simplify checking what is expected as an input/output 😉

lbz/configuration.py
```diff
@@ -47,6 +64,9 @@ def _parse_value(self, value: Any) -> T:
         except Exception as error:
             raise ConfigValueParsingFailed(self._key, value) from error
 
+    def reset(self) -> None:
```
redlickigrzegorz, line 67: Also, I would write one test for this newly added method - we should avoid adding public methods that are not tested, even if they are simple.

tests/conftest.py
```diff
@@ -16,13 +29,41 @@
...
+            "AUTH_REMOVE_PREFIXES": "1",
+            "ALLOWED_PUBLIC_KEYS": json.dumps({"keys": [SAMPLE_PUBLIC_KEY]}),
+            "ALLOWED_AUDIENCES": ",".join(allowed_audiences),
+            "ALLOWED_ISS": "test-issuer",
+            "AWS_LAMBDA_FUNCTION_NAME": "million-dollar-lambda",
+            "EVENTS_BUS_NAME": "million-dollar-lambda-event-bus",
+            "AWS_DEFAULT_REGION": "us-west-2",
+        },
+        clear=True,
+    ):
```
redlickigrzegorz, line 55: As I mentioned in the past, avoid creating multiline context managers - you can always declare a variable above 😉

tests/conftest.py
```diff
@@ -16,13 +29,41 @@
 from lbz.resource import Resource
 from lbz.response import Response
 from lbz.router import Router, add_route
-from tests import SAMPLE_PRIVATE_KEY
+from tests.fixtures.rsa_pair import SAMPLE_PRIVATE_KEY, SAMPLE_PUBLIC_KEY
```
redlickigrzegorz, line 32: If you imported that directly from the "_fixtures_", do we need to import them additionally in `tests/__init__.py`?

tests/conftest.py
```diff
@@ -16,13 +29,41 @@
...
 
 
 @pytest.fixture(scope="session", name="allowed_audiences")
 def allowed_audiences_fixture() -> List[str]:
-    return os.environ["ALLOWED_AUDIENCES"].split(",")
+    return [str(uuid4()), str(uuid4())]
+
+
+@pytest.fixture(autouse=True, name="patch_environ")
+def patch_environ_fixture(allowed_audiences: List[str]) -> Iterator[None]:
```
redlickigrzegorz, line 42: `environ` is patched, that's true, but I would name this fixture according to what is expected, e.g.:
```python
def setting_initial_lbz_configuration(...) -> Iterator[None]:
```

tests/conftest.py
```diff
@@ -16,13 +29,41 @@
...
 from tests.utils import encode_token
 
 
 @pytest.fixture(scope="session", name="allowed_audiences")
 def allowed_audiences_fixture() -> List[str]:
-    return os.environ["ALLOWED_AUDIENCES"].split(",")
+    return [str(uuid4()), str(uuid4())]
+
+
+@pytest.fixture(autouse=True, name="patch_environ")
```
redlickigrzegorz, line 41: You don't have to declare `name` if you don't want to use it below 😉

tests/conftest.py
```diff
@@ -133,7 +174,7 @@ def user_token_fixture(user_cognito: dict) -> str:
     return encode_token(user_cognito)
 
 
-@pytest.fixture(scope="session", name="user")
+@pytest.fixture(name="user")  # scope="session", - TODO: bring that back to reduce run time
```
redlickigrzegorz, line 177: Will we be able to bring that back if the `User` depends on configuration? I think a different local fixture will be required anyway to satisfy the problematic tests.

tests/test_jwt_utils.py
```diff
@@ -78,15 +74,22 @@ def test_missing_correct_audiences(self, caplog: pytest.LogCaptureFixture) -> No
...
-    def test_empty_public_keys(self) -> None:
-        with pytest.raises(RuntimeError):
-            decode_jwt("x")
-
     def test_validate_missing_iss_exception(self) -> None:
         with pytest.raises(SecurityError, match="'exp'"):
             validate_jwt_properties({"allow": "*", "deny": {}})
 
+    @patch.dict(environ, {}, clear=True)
+    def test_empty_public_keys(self) -> None:
```
redlickigrzegorz, line 82: Just out of curiosity, why did you move these two tests?

pdyba: I think they were even removed for a second and then restored - no actual reason at the end.

lbz/jwt_utils.py
```diff
@@ -1,40 +1,27 @@
...
     """
     Checks provided JWT token against allowed tokens.
     """
     try:
         kid_from_jwt_header = jwt.get_unverified_header(auth_jwt_token)["kid"]
-        for key in PUBLIC_KEYS:
+        for key in ALLOWED_PUBLIC_KEYS.value:
             if key["kid"] == kid_from_jwt_header:
-                return key
+                return cast(dict, key)
```
redlickigrzegorz, line 24: When you improve typing, you won't have to cast this variable.

lbz/jwt_utils.py
```diff
@@ -54,29 +41,32 @@ def validate_jwt_properties(decoded_jwt: dict) -> None:
...
 
-    if not ALLOWED_AUDIENCES:
-        raise RuntimeError("ALLOWED_AUDIENCES were not defined")
+    if not ALLOWED_PUBLIC_KEYS.value:
+        raise MissingConfigValue("ALLOWED_PUBLIC_KEYS")
+
+    if not ALLOWED_AUDIENCES.value:
+        raise MissingConfigValue("ALLOWED_AUDIENCES")
+
+    if any("kid" not in public_key for public_key in ALLOWED_PUBLIC_KEYS.value):
```
redlickigrzegorz, line 59: This line checks the correctness of the environment variable, so, I think it should be implemented directly in the parser: `ConfigParser.load_jwt_keys()` 😉

tests/test_jwt_utils.py
```diff
@@ -78,15 +74,22 @@ def test_missing_correct_audiences(self, caplog: pytest.LogCaptureFixture) -> No
...
+        with pytest.raises(MissingConfigValue, match="ALLOWED_PUBLIC_KEYS"):
```
redlickigrzegorz, line 83: This is the message you expect to see:
```suggestion
        with pytest.raises(MissingConfigValue, match="'ALLOWED_PUBLIC_KEYS' was not defined."):
```

tests/test_jwt_utils.py
```diff
@@ -78,15 +74,22 @@ def test_missing_correct_audiences(self, caplog: pytest.LogCaptureFixture) -> No
...
+            decode_jwt("x")
+
+    @patch.dict(
+        environ, {"ALLOWED_PUBLIC_KEYS": json.dumps({"keys": [SAMPLE_PUBLIC_KEY]})}, clear=True
+    )
+    def test_empty_allowed_audiences(self) -> None:
+        with pytest.raises(MissingConfigValue, match="ALLOWED_AUDIENCES"):
```
redlickigrzegorz, line 90: The same here:
```suggestion
        with pytest.raises(MissingConfigValue, match="'ALLOWED_AUDIENCES' was not defined."):
```

tests/test_configuration.py
```diff
@@ -67,11 +70,22 @@ def test_gets_value_from_env_with_json_parser(self) -> None:
...
 
+    @patch.dict(environ, {"RANDOM_VAR": "not-none"})
+    def test_reset_sets_value_to_none_forcing_fetching_it_once_again(self) -> None:
+        env_cfg = EnvValue[str]("RANDOM_VAR")
+        assert env_cfg.value == "not-none"
+
+        env_cfg.reset()
+
+        with patch.dict(environ, {"RANDOM_VAR": "still-not-none"}):
+            env_cfg = EnvValue[str]("RANDOM_VAR")
```
redlickigrzegorz, line 81: You cannot create a completely new variable as this won't confirm the `.reset()` method logic.

tests/test_configuration.py
```diff
@@ -29,14 +29,17 @@ def test_returns_default_when_no_value_set(self) -> None:
...
-        cfg = EnvValue("test")
+        cfg = EnvValue[str]("test")
 
         with pytest.raises(MissingConfigValue, match="'test' was not defined."):
             cfg.value  # pylint: disable=pointless-statement
 
     @patch.object(EnvValue, "getter", MagicMock(return_value=1))
     def test_modifies_config_value_using_declared_parser_function(self) -> None:
-        cfg = EnvValue("RANDOM_VAR", lambda a: a * 10)
+        def times_10_parser(value: int) -> int:
```
redlickigrzegorz, line 39: `value` should be always a string, as it is impossible to get something else from env:
```suggestion
        def times_10_parser(value: str) -> int:
```

tests/test_configuration.py
```diff
@@ -29,14 +29,17 @@ def test_returns_default_when_no_value_set(self) -> None:
         assert cfg.value == 42
 
     def test_raises_missing_configuration_when_value_is_none(self) -> None:
-        cfg = EnvValue("test")
+        cfg = EnvValue[str]("test")
 
         with pytest.raises(MissingConfigValue, match="'test' was not defined."):
             cfg.value  # pylint: disable=pointless-statement
 
     @patch.object(EnvValue, "getter", MagicMock(return_value=1))
```
redlickigrzegorz, line 37: To make it more correct you should patch environ instead:
```suggestion
    @patch.dict(environ, {"RANDOM_VAR": "1"})
```

tests/test_configuration.py
```diff
@@ -29,14 +29,17 @@ def test_returns_default_when_no_value_set(self) -> None:
...
     def test_modifies_config_value_using_declared_parser_function(self) -> None:
-        cfg = EnvValue("RANDOM_VAR", lambda a: a * 10)
+        def times_10_parser(value: int) -> int:
+            return value * 10
+
+        cfg = EnvValue[int]("RANDOM_VAR", times_10_parser)
```
redlickigrzegorz, line 42:
```suggestion
        cfg = EnvValue("RANDOM_VAR", times_10_parser)
```

tests/test_configuration.py
```diff
@@ -80,8 +94,34 @@ class TestSSMConfig:
...
-        cfg = SSMValue("key_name")
+        cfg = SSMValue[str]("key_name")
 
         assert cfg.value == "test_value"
 
         mocked_get_parameter.assert_called_once_with("key_name")
+
+
+class TestConfigParser:
+    def test_split_by_comma_returns_list(self) -> None:
```
redlickigrzegorz, line 105:
```suggestion
    def test__split_by_comma__returns_list(self) -> None:
```

tests/test_configuration.py
```diff
@@ -80,8 +94,34 @@ class TestSSMConfig:
...
+            ("TruE", True),
+            ("0", False),
+            ("False", False),
+            ("fLase", False),
+            ("not-true", False),
+            ("N/A", False),
+            ("xxxxx", False),
+        ],
+    )
+    def test_cast_to_bool_returns_bool(self, input_value: str, expected_value: bool) -> None:
```
redlickigrzegorz, line 123:
```suggestion
    def test__cast_to_bool__returns_bool(self, input_value: str, expected_value: bool) -> None:
```

tests/test_configuration.py
```diff
@@ -80,8 +94,34 @@ class TestSSMConfig:
...
+        assert ConfigParser.cast_to_bool(input_value) == expected_value
+
+    def test_load_jwt_keys_return_value_of_keys(self) -> None:
```
redlickigrzegorz, line 126:
```suggestion
    def test__load_jwt_keys__return_value_of_keys(self) -> None:
```

redlickigrzegorz (approved):

lbz/_cfg.py
```diff
@@ -0,0 +1,23 @@
...
+
+# LBZ configuration
+LBZ_DEBUG_MODE = EnvValue[bool]("LBZ_DEBUG_MODE", default=False, parser=ConfigParser.cast_to_bool)
+LOGGING_LEVEL = EnvValue("LOGGING_LEVEL", default="INFO")
+CORS_HEADERS = EnvValue[List[str]]("CORS_HEADERS", default=[], parser=ConfigParser.split_by_comma)
+CORS_ORIGIN = EnvValue[List[str]]("CORS_ORIGIN", default=[], parser=ConfigParser.split_by_comma)
+
+# AWS related configuration
+AWS_LAMBDA_FUNCTION_NAME = EnvValue[str]("AWS_LAMBDA_FUNCTION_NAME")
+EVENTS_BUS_NAME = EnvValue[str]("EVENTS_BUS_NAME")
```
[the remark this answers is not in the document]
pdyba: Actually all of the EnvValue started to expect that type definition after recent change - mypy is not happy if the don't have it.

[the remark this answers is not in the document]
pdyba: When there is no default value or parser defined ;/

PR #71, opened 2022-10-06
Unifies Based Broker for LambdaApi and EventAPI

pdyba:


lbz/events/broker.py
```diff
@@ -2,23 +2,26 @@
...
 
 logger = get_logger(__name__)
 
 
 # TODO: type_key and data_key will be const for EventBridge and different set for Cognito Events
-class EventBroker:
+class EventBroker(BaseBroker):
     def __init__(
         self,
         mapper: Dict[str, List[Callable]],
```
redlickigrzegorz, line 15: We can make this type more precise I think:
```suggestion
        mapper: Dict[str, List[Callable[[Event], None]]],
```

lbz/handlers.py
```diff
@@ -8,12 +8,21 @@
 T = TypeVar("T")
 
 
-class BaseHandler(Generic[T], metaclass=ABCMeta):
+class BaseBroker(Generic[T], metaclass=ABCMeta):
+    def __init__(
+        self,
+        event: dict,
+        context: object = None,
+    ) -> None:
```
redlickigrzegorz, line 16:
```suggestion
    def __init__(self, event: dict, context: object = None) -> None:
```

redlickigrzegorz (approved):

lbz/handlers.py
```diff
@@ -9,11 +9,16 @@
 
 
 class BaseHandler(Generic[T], metaclass=ABCMeta):
+    def __init__(self, event: dict, context: object = None) -> None:
```
redlickigrzegorz, line 12: Now, when I look at this I think `context` should not be optional 🤔 
```suggestion
    def __init__(self, event: dict, context: object) -> None:
```

redlickigrzegorz: If we keep it like this, we will have to add an `if` condition every time when using this value 😐

lbz/events/broker.py
```diff
@@ -2,23 +2,26 @@
...
 # TODO: type_key and data_key will be const for EventBridge and different set for Cognito Events
-class EventBroker:
+class EventBroker(BaseHandler):
     def __init__(
         self,
-        mapper: Dict[str, List[Callable]],
-        raw_event: dict,
+        mapper: Dict[str, List[Callable[[Event], None]]],
+        event: dict,
+        context: object = None,
```
redlickigrzegorz, line 17: According to [#989723417](https://github.com/pdyba/lambdalizator/pull/71#discussion_r989723417), `context` cannot be optional.

lbz/lambdas/broker.py
```diff
@@ -16,20 +16,19 @@ def __init__(
         event: dict,
         context: object = None,
```
redlickigrzegorz, line 17: According to [#989723417](https://github.com/pdyba/lambdalizator/pull/71#discussion_r989723417), `context` cannot be optional.

lbz/events/broker.py
```diff
@@ -2,23 +2,26 @@
...
 from lbz.events.event import Event
+from lbz.handlers import BaseHandler
 from lbz.misc import get_logger
 
 logger = get_logger(__name__)
 
 
 # TODO: type_key and data_key will be const for EventBridge and different set for Cognito Events
-class EventBroker:
+class EventBroker(BaseHandler):
```
redlickigrzegorz, line 12: You should remove `pre_handle()` and `post_handle()` from this class as it started using `BaseHandler`.

I don't know why I didn't see that before 🤦

PR #80, opened 2023-03-09
Adds CognitoEventBroker improves Event API

pdyba:


lbz/events/enums.py
```diff
@@ -0,0 +1,21 @@
...
+
+    EMAIL_FORGOT_PASSWORD = "CustomEmailSender_ForgotPassword"
+    EMAIL_RESEND_CODE = "CustomEmailSender_ResendCode"
+    EMAIL_SIGNUP = "CustomEmailSender_SignUp"
+    EMAIL_VERIFY_USER_ATTRIBUTE = "CustomEmailSender_VerifyUserAttribute"
+    EMAIL_UPDATE_USER_ATTRIBUTE = "CustomEmailSender_UpdateUserAttribute"
+    EMAIL_ADMIN_CREATEUSER = "CustomEmailSender_AdminCreateUser"
+    EMAIL_ACCOUNT_TAKE_OVER_NOTIFICATION = "CustomEmailSender_AccountTakeOverNotification"
+
+    SMS_UPDATE_USER_ATTRIBUTE = "CustomSMSSender_UpdateUserAttribute"
```
redlickigrzegorz, line 15: Can we put it in the order declared by AWS or at least alphabetically?

lbz/events/api.py
```diff
@@ -29,7 +29,7 @@ def __init__(self) -> None:
 
     def __repr__(self) -> str:
         return (
-            f"<EventAPI bus: {self._bus_name} Events: pending={len(self._pending_events)} "
+            f"<EventAPI bus: {self.bus_name} Events: pending={len(self._pending_events)} "
```
redlickigrzegorz, line 32: I don't think you need to use the public property when you are inside the class 😉

lbz/events/api.py
```diff
@@ -46,6 +46,10 @@ def set_bus_name(self, bus_name: str) -> None:
     def sent_events(self) -> List[Event]:
         return deepcopy(self._sent_events)
 
+    @property
+    def bus_name(self) -> str:
```
redlickigrzegorz, line 50: I don't mind adding this property but it won't help us solve an issue of sending events to custom event bus on demand 🤷 

As I mentioned outside GitHub, I think, we should rather consider extending the `@event_emitter` decorator and at least talk about it before implementing.

redlickigrzegorz, line 50: Also, it would be nice to declare it somewhere higher, not between `sent_events` and `pending_events` 😉

redlickigrzegorz: As we agreed, the issue will be handled, at least temporarily, in the outside code.

CHANGELOG.md
```diff
@@ -306,8 +306,15 @@ Released 2022-11-03
 - Adds examples for Event aware API, Event handler (broker), and stand-alone Event producer
 - Adds examples for Lambda client and broker. 
 
+### Version 0.5.17
+Released 2023-03-10
+
+- Adds Cognito Event Broker
+- Adds Cognito generated event enum.
+-
```
redlickigrzegorz, line 314: Point without text:
```suggestion
```

lbz/events/enums.py
```diff
@@ -0,0 +1,17 @@
+class CognitoEventType:
+    # docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-email-sender.html
```
redlickigrzegorz, line 2: It needs to be a full URL, otherwise, e.g. PyCharm does not recognize it:
```suggestion
    # https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-email-sender.html
```

lbz/events/enums.py
```diff
@@ -0,0 +1,17 @@
...
+    EMAIL_ACCOUNT_TAKE_OVER_NOTIFICATION = "CustomEmailSender_AccountTakeOverNotification"
+    EMAIL_ADMIN_CREATE_USER = "CustomEmailSender_AdminCreateUser"
+    EMAIL_FORGOT_PASSWORD = "CustomEmailSender_ForgotPassword"
+    EMAIL_RESEND_CODE = "CustomEmailSender_ResendCode"
+    EMAIL_SIGNUP = "CustomEmailSender_SignUp"
+    EMAIL_UPDATE_USER_ATTRIBUTE = "CustomEmailSender_UpdateUserAttribute"
+    EMAIL_VERIFY_USER_ATTRIBUTE = "CustomEmailSender_VerifyUserAttribute"
+    # docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-sms-sender.html
```
redlickigrzegorz, line 10: The same here:
```suggestion

    # https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-sms-sender.html
```

lbz/events/broker.py
```diff
@@ -8,17 +8,21 @@
...
 
+EVENT_BRIDGE_DEFAULT_TYPE_KEY = "detail-type"
+EVENT_BRIDGE_DEFAULT_DATA_KEY = "detail"
+COGNITO_DEFAULT_TYPE_KEY = "triggerSource"
+COGNITO_DEFAULT_DATA_KEY = "request"
 
-# TODO: type_key and data_key will be const for EventBridge and different set for Cognito Events
-class EventBroker(BaseHandler[None]):
+
+class BaseEventBroker(BaseHandler[None]):
```
redlickigrzegorz, line 17: Once you created this class, most likely all the old tests should start using it instead 😉 

And we will need only one test for `EventBroker` exactly as you did for `CognitoEventBroker`.

lbz/events/broker.py
```diff
@@ -40,3 +44,55 @@ def _get_handlers(self) -> List[Callable]:
...
+
+class EventBroker(BaseEventBroker):
+    def __init__(
+        self,
+        mapper: Mapping[str, List[Callable[[Event], None]]],
+        event: dict,
+        context: LambdaContext,
+        *,
+        type_key: str = EVENT_BRIDGE_DEFAULT_TYPE_KEY,
+        data_key: str = EVENT_BRIDGE_DEFAULT_DATA_KEY,
```
redlickigrzegorz, line 57: Let's remove these arguments to make `EventBroker` our standard class as it was agreed in the past 😉 

Of course, this won't be backward compatible but if really someone is using these arguments, they will simply add a `Base` prefix before the name of the class 😉

lbz/events/broker.py
```diff
@@ -40,3 +44,55 @@ def _get_handlers(self) -> List[Callable]:
...
+        super().__init__(
+            mapper,
+            event,
+            context,
+            type_key=type_key,
+            data_key=data_key,
+        )
+
+
+class EventBridgeBroker(BaseEventBroker):
```
redlickigrzegorz, line 68: According to the above comment, this class will become redundant - we can remove it.

lbz/events/broker.py
```diff
@@ -8,17 +8,21 @@
 
 logger = get_logger(__name__)
 
+EVENT_BRIDGE_DEFAULT_TYPE_KEY = "detail-type"
+EVENT_BRIDGE_DEFAULT_DATA_KEY = "detail"
+COGNITO_DEFAULT_TYPE_KEY = "triggerSource"
+COGNITO_DEFAULT_DATA_KEY = "request"
```
redlickigrzegorz, line 14: I don't think we should create global variables for these strings, it gives the ability to manipulate them, but also they will only be used once:
```suggestion
```

tests/test_events/test_broker.py
```diff
@@ -83,8 +100,42 @@ def handler(event: Event) -> None:
...
+        func_2.assert_called_once_with(expected_event)
+
+
+class TestEventBroker:
+    def test_broker_works_properly(self) -> None:
+        func_1 = MagicMock()
+        func_2 = MagicMock()
+        expected_event = Event({"y": 1}, event_type="x")
+        mapper = {"x": [func_1, func_2]}
+        event = {"detail-type": "x", "detail": {"y": 1}, "userName": "usr-123"}
```
redlickigrzegorz, line 136:
```suggestion
        event = {"detail-type": "x", "detail": {"y": 1}}
```

redlickigrzegorz (approved):

PR #83, opened 2023-09-19
Adjust behaviour of Event API and related

pdyba:


lbz/events/api.py
```diff
@@ -97,6 +94,7 @@ def _create_eb_entry(self, new_event: Event) -> PutEventsRequestEntryTypeDef:
 
 def event_emitter(function: Callable) -> Callable:
     """Decorator that makes function an emitter - automatically sends pending events on success"""
+    EventAPI().clear()
```
redlickigrzegorz, line 97: What about the below clearing? 🤔 Of course, we should get rid of all the pending events but the current logic says that failed/sent events should be kept during the whole "session".

lbz/resource.py
```diff
@@ -239,6 +239,7 @@ class EventAwareResource(Resource):
     def __init__(self, event: dict):
         super().__init__(event)
         self.event_api = EventAPI()
+        self.event_api.clear()
```
redlickigrzegorz, line 242: I have exactly the same comment here, I don't think we should clear all the lists if the response is not OK 🤔 

The `clear()` method should be used only in two places OR nowhere if we are fully confident that our Singleton works correctly in Lambda functions.

tests/test_events/test_api.py
```diff
@@ -199,15 +199,31 @@ def test_singleton_pattern_working_correctly_for_event_api(self, mock_send: Magi
         )
 
     @patch.object(Boto3Client, "eventbridge", MagicMock())
-    def test_second_send_clears_everything(self) -> None:
+    def test_second_send_does_not_clear_everything(self) -> None:
```
redlickigrzegorz, line 202: That is not exactly what you check here, you test that events are left in lists. This is quite important as someone in the future won't know why we connect clearing with the send logic 😉

tests/test_events/test_api.py
```diff
@@ -199,15 +199,31 @@ def test_singleton_pattern_working_correctly_for_event_api(self, mock_send: Magi
...
         self.event_api.send()
         self.event_api.send()
 
         assert self.event_api.failed_events == []
-        assert self.event_api.sent_events == []
+        assert self.event_api.sent_events == [event]
+        assert self.event_api.pending_events == []
+
+    @patch.object(Boto3Client, "eventbridge", MagicMock())
+    def test_conclusive_send_extends_event_list(self) -> None:
```
redlickigrzegorz, line 214: This is more or less the same test as what is above but with more events 🤷

redlickigrzegorz, line 214: It would be good to test it also with the failed events as the second `.send()`, even if fully correct, will raise an error.

I am not 100% sure about this logic but that is how behaves after your changes 🤔

tests/test_events/test_api.py
```diff
@@ -260,3 +276,16 @@ def decorated_function() -> None:
         assert not EventAPI().sent_events
         assert not EventAPI().pending_events
         assert not EventAPI().failed_events
+
+    def test_clears_que_when_wrapping_without_calling(self) -> None:
```
redlickigrzegorz, line 280: You have typo here but also it would be good to describe what is expected:
```suggestion
    def test_always_clears_queues_before_actually_decorating_function(self) -> None:
```

tests/test_resource.py
```diff
@@ -484,8 +484,8 @@ def test_method(self) -> Response:
 
         XResource(event)()
 
+        mocked_event_api_clear.assert_called_once()
```
redlickigrzegorz, line 487: This is not an ideal test as the order of running these methods matters 😐 

I think we need to resign from this way of mocking and test it more directly by checking the list of events.

lbz/events/api.py
redlickigrzegorz: It looks like you updated the `EventAwareResource` class but not the `event_emitter` decorator.

tests/test_events/test_api.py
redlickigrzegorz: You skipped the above comment ☝️

tests/test_events/test_api.py
```diff
@@ -223,6 +239,48 @@ def test_clear(self) -> None:
         assert self.event_api.sent_events == []
         assert self.event_api.pending_events == []
 
+    @patch.object(Boto3Client, "eventbridge", MagicMock())
+    def test_clear_pending_clears_only_pending_events(self) -> None:
```
redlickigrzegorz, line 243: Please, stay consistent with the naming convention that we agreed to follow in all repositories.

tests/test_events/test_api.py
```diff
@@ -223,6 +239,48 @@ def test_clear(self) -> None:
...
+
+    @patch.object(Boto3Client, "eventbridge")
+    def test_clear_failed_clears_only_failed_events(self, mock_send: MagicMock) -> None:
+        assert self.event_api.failed_events == []
+        mock_send.put_events.side_effect = NotADirectoryError
+        event = MyTestEvent({"x": 1})
+        self.event_api.register(event)
+        with pytest.raises(RuntimeError):
+            self.event_api.send()
+        assert self.event_api.failed_events == [event]
```
redlickigrzegorz, line 276: You don't need these additional assertions, this test checks completely something else than storing an event on the list of failed events 😉

tests/test_events/test_api.py
```diff
@@ -223,6 +239,48 @@ def test_clear(self) -> None:
...
+        self.event_api.register(event)
+
+        self.event_api.clear_failed()
+
+        assert self.event_api.failed_events == []
+        assert self.event_api.pending_events == [event]
```
redlickigrzegorz, line 282: You skipped one list here.

redlickigrzegorz (commented):
Please, go through all of the comments that you skipped - the current version is incomplete 😉

tests/test_events/test_api.py
redlickigrzegorz: You skipped the above comment ☝️

pdyba: Not sure what You want me to do here then.

redlickigrzegorz: I mean `clear` has nothing to do with `send` after your changes and putting it into the test names is misleading.

But I will write more comments about it as the current version of tests is even more messy 🤷

tests/test_resource.py
```diff
@@ -485,12 +485,14 @@ def test_method(self) -> Response:
         XResource(event)()
 
         mocked_event_api_send.assert_called_once()
-        mocked_event_api_clear.assert_not_called()
+        assert EventAPI().sent_events == []
+        assert EventAPI().pending_events == []
+        assert EventAPI().failed_events == []
```
redlickigrzegorz, line 490: These two tests do not emit any events, so, it is obvious that these lists are empty from the beginning to the very end 😉 

If you really don't want to remove mocks, I have another solution on how to test this logic fully (not beautiful but at least, it confirms the flow):

```python
class TestEventAwareResource:
    def test_initializes_completely_new_event_api_when_building_resource(self) -> None:
        class XResource(EventAwareResource):
            pass

        with patch.object(EventAPI, "clear", autospec=True) as mocked_clear:
            resource = XResource(event)

        assert resource.event_api is EventAPI()
        mocked_clear.assert_called_once_with(resource.event_api)

    def test_emits_all_pending_events_when_request_is_successfully_handled(self) -> None:
        class XResource(EventAwareResource):
            @add_route("/")
            def test_method(self) -> Response:
                return Response({"message": "x"})

        resource = XResource(event)
        with patch.object(resource, "event_api", autospec=True) as mocked_event_api:
            resource()

        mocked_event_api.send.assert_called_once_with()
        mocked_event_api.clear.assert_not_called()
        mocked_event_api.clear_pending.assert_not_called()

    def test_clears_pending_events_when_request_failed_during_handling(self) -> None:
        class XResource(EventAwareResource):
            @add_route("/")
            def test_method(self) -> None:
                raise TypeError

        resource = XResource(event)
        with patch.object(resource, "event_api", autospec=True) as mocked_event_api:
            resource()

        mocked_event_api.send.assert_not_called()
        mocked_event_api.clear.assert_not_called()
        mocked_event_api.clear_pending.assert_called_once_with()
```

tests/test_events/test_api.py
```diff
@@ -198,16 +198,36 @@ def test_singleton_pattern_working_correctly_for_event_api(self, mock_send: Magi
             ]
         )
 
-    @patch.object(Boto3Client, "eventbridge", MagicMock())
-    def test_second_send_clears_everything(self) -> None:
+    @patch.object(Boto3Client, "eventbridge")
+    def test_second_send_does_not_clear_everything(self, mock_send: MagicMock) -> None:
```
redlickigrzegorz, line 202: It looks like you mixed up tests with their names 😉 
```suggestion
    def test__send__raises_error_only_when_particular_attempt_failed(
        self, mock_send: MagicMock
    ) -> None:
```

tests/test_events/test_api.py
```diff
@@ -198,16 +198,36 @@ def test_singleton_pattern_working_correctly_for_event_api(self, mock_send: Magi
...
+        mock_send.put_events.side_effect = NotADirectoryError
         event = MyTestEvent({"x": 1})
         self.event_api.register(event)
+        with pytest.raises(RuntimeError):
+            self.event_api.send()
+        mock_send.put_events.side_effect = None
```
redlickigrzegorz, line 208: Also, you can declare the side effect at the beginning:
```suggestion
        mock_send.put_events.side_effect = [NotADirectoryError, None]
        event = MyTestEvent({"x": 1})

        self.event_api.register(event)
        with pytest.raises(RuntimeError):
            self.event_api.send()
```

tests/test_events/test_api.py
```diff
@@ -198,16 +198,36 @@ def test_singleton_pattern_working_correctly_for_event_api(self, mock_send: Magi
...
+        self.event_api.register(event)
+        self.event_api.register(event)
+        self.event_api.send()
 
+        assert self.event_api.failed_events == [event]
+        assert self.event_api.sent_events == [event, event]
+        assert self.event_api.pending_events == []
+
+    @patch.object(Boto3Client, "eventbridge", MagicMock())
+    def test_raises_only_when_failed_in_current_send_scope(self) -> None:
```
redlickigrzegorz, line 218:
```suggestion
    def test__send__continuously_extends_lists_of_events_during_next_attempts(self) -> None:
```

tests/test_events/test_api.py
```diff
@@ -248,15 +309,32 @@ def decorated_function() -> None:
...
-    def test_clears_queues_when_error_appeared_during_running_decorated_function(self) -> None:
+    def test_clears_pending_queue_when_error_appeared_during_running_decorated_function(
+        self,
+    ) -> None:
         @event_emitter
         def decorated_function() -> None:
             EventAPI().register(MyTestEvent({"x": 1}))
             raise RuntimeError
 
+        EventAPI().register(MyTestEvent({"x": 1}))
```
redlickigrzegorz, line 320: Let's differentiate it to make it more obvious:
```suggestion
        EventAPI().register(MyTestEvent({"x": 2}))
```

tests/test_events/test_api.py
```diff
@@ -248,15 +309,32 @@ def decorated_function() -> None:
...
+        EventAPI().send()
         with pytest.raises(RuntimeError):
             decorated_function()
 
+        assert EventAPI().sent_events == [MyTestEvent({"x": 1})]
```
redlickigrzegorz, line 325:
```suggestion
        assert EventAPI().sent_events == [MyTestEvent({"x": 2})]
```

redlickigrzegorz (approved):

PR #84, opened 2023-10-26
Rest over lambda (with API GW skip) interface

pdyba:

PR #86, opened 2024-01-02
Adjust the code to the latest standards set by Python 3.12

redlickigrzegorz:
This pull request is the continuation of the work started in #85 but it focuses mostly on Python 3.12 😎 

Unfortunately, not everything is compatible with Python 3.8, thus starting from now, Lambdalizator requires at least 3.9.


andreistr21 (commented):
Merged "Adjust the code to the latest standards set by Python 3.12" from feature/python-3.12 to master

andreistr21:
multidict package does not support python 3.12 yet.

lbz/misc.py
```diff
@@ -24,12 +22,6 @@ def __getitem__(self, key: str) -> Any:
 
 
 class Singleton(type):
-    """
```
pdyba, line 27: Why remove it - this specific usage is not 100% obvious.

lbz/router.py
```diff
@@ -1,17 +1,14 @@
-"""
-Router module.
-"""
+"""Router module."""
```
pdyba, line 1: I would consider removing that and Router class below.

pdyba:
> multidict package does not support python 3.12 yet.

It's almost stale package, I woudn't worry about that. And with API GW 2.0 Event we could consider dropping it.

pdyba:
Also as You are updating it to 0.6.0 You should resolve those two todos.
Screenshot 2024-01-03 at 13 55 55
Screenshot 2024-01-03 at 13 55 43

redlickigrzegorz:
> multidict package does not support python 3.12 yet.

As long as it can be installed and works correctly (according to our tests), we should not worry that much.

But definitely, this will be monitored in the upcoming days/weeks/months 🙂

lbz/misc.py
redlickigrzegorz: I would say the declaration of this class says enough but ok, I can restore it 👍 

Removed only because of missing first line of the docstring which is required now.

lbz/router.py
redlickigrzegorz: I thought about removing docstrings in many places as they were soooo obvious, will remove them 😌

pdyba:
> > Also as You are updating it to 0.6.0 You should resolve those two todos.
> > Screenshot 2024-01-03 at 13 55 55 Screenshot 2024-01-03 at 13 55 43
> 
> The first one is coming from your code (#84) - I don't have it yet.
> 
> But this pull request is not intended to change code - we should either:
> 
> 1. Create one big v0.6 package with all required things inside (multiple pull requests).
> 2. Resolve these TODOs when we will work on Lambdalizator v0.7.
> 
> As for me, option 2. is better as it won't block us.

Then please adjust the deprecation warning to 0.7.0 - option 2.

pdyba (approved):

PR #101, opened 2024-05-16
Extra data for errors

pdyba:
Allows extending erros and excetpion responses with extra data from outside when needed


lbz/exceptions.py
```diff
@@ -20,14 +20,19 @@ def __init__(self, key: str, value: Any) -> None:
 
 
 class LambdaFWException(Exception):
-    """Standardised for AWS Lambda exception class."""
+    """Standardised for AWS Lambda exception class.
+
+    extra allows extending the response with additional data when needed"""
```
andreistr21, line 25:
```suggestion
    extra allows extending the response with additional data when needed.
    """
```

Makefile
```diff
@@ -99,7 +99,7 @@ bandit:
 pip-audit:
 	pip-audit --version
 	# TODO: Fix the issue with the vulnerable ecdsa library
-	pip-audit -r requirements.txt --ignore-vuln GHSA-wj6h-64fc-37mp
+	pip-audit -r requirements.txt --ignore-vuln GHSA-wj6h-64fc-37mp --ignore-vuln GHSA-cjwg-qfpm-7377 --ignore-vuln GHSA-6c5p-j8vq-pqhj
```
andreistr21, line 102: You should update the TODO comment too

andreistr21 (approved):

tests/test_exceptions.py
```diff
@@ -28,12 +28,13 @@ class TestException(LambdaFWException):
         status_code = 444
         error_code = "TEST_ERR"
 
-    exception = TestException()
+    exception = TestException(extra={"extra": "data"})
```
redlickigrzegorz, line 31: Also, having a key named extra is rather misleading.

tests/test_response.py
```diff
@@ -122,6 +122,24 @@ class RandomException(LambdaFWException):
             "statusCode": 500,
         }
 
+    def test__from_exception__adds_extra_data_to_response_body_if_only_declared(self) -> None:
+        class RandomException(LambdaFWException):
+            error_code = "RAND001"
+
+        response = Response.from_exception(RandomException(extra={"extra": "random"}), "req-id")
```
redlickigrzegorz, line 129: The same here, do not use a key named `extra` as it may be wrongly interpreted during reading the results of this test.

tests/test_response.py
```diff
@@ -122,6 +122,24 @@ class RandomException(LambdaFWException):
...
+
+        assert response.to_dict() == {
+            "body": (
+                '{"message":"Server got itself in trouble",'
```
redlickigrzegorz, line 133: It would be better to compare `body` separately making a dictionary from that first 😉 

Thanks to that, you will avoid order issues.

lbz/exceptions.py
```diff
@@ -26,12 +26,13 @@ class LambdaFWException(Exception):
...
-        super().__init__(message)
-        if message:
-            self.message = message
-        if error_code:
-            self.error_code = error_code
+    def __init__(
+        self, message: str = "", error_code: str | None = None, extra: dict | None = None
+    ) -> None:
+        self.message = message or self.message
+        super().__init__(self.message)
```
redlickigrzegorz, line 33: There should be nothing before super initialization - inheritance is based on extending the base class 😉 
```suggestion
        super().__init__(message)

        self.message = message or self.message
```

tests/test_exceptions.py
```diff
@@ -34,6 +34,7 @@ class TestException(LambdaFWException):
     assert exception.message == "No error message"
     assert exception.error_code == "TEST_ERR"
     assert exception.status_code == 444
+    assert exception.extra == {}
 
 
 def test__lambda_fw_exception__respects_values_provided_during_initialization() -> None:
```
redlickigrzegorz, line 40: As I mentioned [here](https://github.com/pdyba/lambdalizator/pull/101#discussion_r1605810800), you should adjust this test to show the new argument works 😉

tests/test_response.py
```diff
@@ -122,6 +123,27 @@ class RandomException(LambdaFWException):
...
+            RandomException(extra={"added_data": "random"}), "req-id"
+        )
+
+        assert response.to_dict() == {
+            "body": ANY,
+            "headers": {"Content-Type": ContentType.JSON},
+            "isBase64Encoded": False,
+            "statusCode": 500,
+        }
+        assert response.body == {
```
redlickigrzegorz, line 140: Hmmm, if you didn't plan to use value from `.to_dict()`, you can compare all the attributes this way 😉

redlickigrzegorz, line 140: Hmmm, if you didn't plan to use the value from `.to_dict()`, you could simply compare all the attributes this way 😉

redlickigrzegorz (approved):

PR #102, opened 2024-11-13
Adds WebSocket API GW request

pdyba:
cleans up requests and tests
start depracating Request in favour of HTTPRequest


lbz/request.py
```diff
@@ -84,10 +103,48 @@ def json_body(self) -> dict | None:
...
 
-    def to_dict(self) -> dict:
-        copied = {k: v for k, v in self.__dict__.items() if not k.startswith("_")}
-        copied["headers"] = dict(copied["headers"])
-        copied["user"] = repr(copied["user"])
-        if copied["query_params"] is not None:
-            copied["query_params"] = dict(copied["query_params"])
-        return copied
+
+class WebSocketRequest(BaseRequest):
```
redlickigrzegorz, line 107: Why did you implement this class here, once you created a new dedicated package for websocket things? 🤔

lbz/request.py
```diff
@@ -84,10 +103,48 @@ def json_body(self) -> dict | None:
...
+    """Represents request from Web Socket Secure API Gateway."""
```
redlickigrzegorz, line 108: Maybe we can create a similar Event class to the one we have for REST? It should simplify the work with it 😉

redlickigrzegorz, line 108: Also, adding the documentation link would be helpful. [This is it](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-mapping-template-reference.html), right?

lbz/request.py
```diff
@@ -84,10 +103,48 @@ def json_body(self) -> dict | None:
...
+        request_details: dict,
+        context: dict,
+        is_base64_encoded: bool,
+        user: User | None = None,
+        headers: CIMultiDict | None = None,
+    ):
+        self.headers = headers
+        self.context = context
+        self.user = user
+        self.action = request_details.pop("routeKey")
```
redlickigrzegorz, line 123: It would be nice to confirm the documentation that you used as I see that all these keys are under `context` 🤷

lbz/request.py
```diff
@@ -68,6 +50,43 @@ def _safe_json_loads(payload: str | bytes) -> Any:
         except ValueError as error:
             raise BadRequestError(f"Invalid payload.\nPayload body:\n {repr(payload)}") from error
 
+
+class Request(BaseRequest):
```
redlickigrzegorz, line 54: If you already changed the version to 0.7.0, you can rename this class instead of creating the alias below 😉

redlickigrzegorz, line 54: Also, it can finally be moved to the `rest` package.

lbz/request.py
```diff
@@ -68,6 +50,43 @@ def _safe_json_loads(payload: str | bytes) -> Any:
...
+        self.method = method
+        self.context = context
+        self.stage_vars = stage_vars
+        self.user = user
+        super().__init__(body=body, is_base64_encoded=is_base64_encoded)
+
+    def __repr__(self) -> str:
+        return f"<Request {self.method} >"
+
+    def to_dict(self) -> dict:
```
redlickigrzegorz, line 82: As we agreed a while ago, all the properties should be implemented before methods like this (original order).

lbz/request.py
```diff
@@ -84,10 +103,48 @@ def json_body(self) -> dict | None:
...
+        return f"<Request {self.action_type} - {self.action} >"
+
+    def is_connection_request(self) -> bool:
+        return self.action_type is ActionType.CONNECT
+
+    def is_disconnection_request(self) -> bool:
+        return self.action_type is ActionType.DISCONNECT
+
+    @property
+    def json_body(self) -> dict | None:
```
redlickigrzegorz, line 141: I started thinking whether these two classes are really that similar to inherit from the same base one 🤔

tests/test_request.py
```diff
@@ -128,3 +123,87 @@ def test_to_dict(self, sample_request_with_user: Request) -> None:
...
+            (ActionType.CONNECT, True),
+            (ActionType.DISCONNECT, False),
+            (ActionType.MESSAGE, False),
+        ],
+    )
+    def test__is_connection_request__checks_if_request_was_a_connection_request(
+        self, action_type: str, is_connection_req: bool
+    ) -> None:
+        request = WebSocketRequest(
+            "",
```
redlickigrzegorz, line 141: Please, use the keywords arguments, otherwise, it is hard to see what you want to pass here.

tests/test_request.py
```diff
@@ -128,3 +123,87 @@ def test_to_dict(self, sample_request_with_user: Request) -> None:
...
+                "domainName": "xxx.com",
+                "stage": "prod",
+            },
+            {},
+            False,
+        )
+
+        assert request.is_disconnection_request() == is_connection_req
+
+    def test_json_body_dict(self) -> None:
```
redlickigrzegorz, line 181: Follow the naming convention as you did above.

tests/test_request.py
```diff
@@ -128,3 +123,87 @@ def test_to_dict(self, sample_request_with_user: Request) -> None:
...
+                "connectionId": "aaa123",
+                "domainName": "xxx.com",
+                "stage": "prod",
+            },
+            {},
+            False,
+        )
+        assert request.json_body == {"x": "t1"}
+
+    def test_json_body_json(self) -> None:
```
redlickigrzegorz, line 196: ditto

requirements-dev.txt
```diff
@@ -107,84 +103,83 @@ packaging==23.2
...
     #   pip-audit
-rich==13.7.0
+rich==13.9.4
     # via
     #   bandit
     #   pip-audit
 six==1.16.0
     # via
-    #   -c requirements.txt
+    #   -c /Users/dybapi/PycharmProjects/lambdalizator/requirements.txt
```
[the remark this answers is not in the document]
redlickigrzegorz: As for me, you can move bumping packages to a separate PR that should be rather easy to review/merge 😉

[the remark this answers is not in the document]
pdyba: fixed - I did not expect that kind of issues

lbz/request.py
pdyba: The Event class will probably be in separated request hopefully it can be the same one

lbz/request.py
pdyba: I've run the code ;) I can share the actual payloads - they will be there in future PRs (around Event)

lbz/request.py
pdyba: The remaining methods in BaseRequest makes sense to have shared ancestor. It might still change before we actually release 0.7.0

lbz/request.py
redlickigrzegorz: Based on the documentation, I doubt it can be the same but I will judge it when I see 😉

lbz/request.py
redlickigrzegorz: Ok, so, I will verify it then 👍

lbz/request.py
redlickigrzegorz: Ok, we will return to this comment in the next pull requests 👍

lbz/request.py
redlickigrzegorz: FYI: #105

lbz/request.py
redlickigrzegorz: FYI: #105

lbz/request.py
redlickigrzegorz: FYI: #105

lbz/websocket/request.py
```diff
@@ -0,0 +1,56 @@
...
+        self.headers = headers
+        self.context = context
+        self.user = user
+        self.action = request_details.pop("routeKey")
+        self.action_type = request_details.pop("eventType")
+        self.connection_id = request_details.pop("connectionId")
+        self.domain = request_details.pop("domainName")
+        self.stage = request_details.pop("stage")
+        self.details = request_details
+        super().__init__(body=body, is_base64_encoded=is_base64_encoded)
```
redlickigrzegorz, line 38: The same here, it should be done as a first operation:
```suggestion
        super().__init__(body=body, is_base64_encoded=is_base64_encoded)
        self.headers = headers
        self.context = context
        self.user = user
        self.action = request_details.pop("routeKey")
        self.action_type = request_details.pop("eventType")
        self.connection_id = request_details.pop("connectionId")
        self.domain = request_details.pop("domainName")
        self.stage = request_details.pop("stage")
        self.details = request_details
```

lbz/websocket/request.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    def __init__(
+        self,
+        body: str | bytes | dict,
+        request_details: dict,
+        context: dict,
+        is_base64_encoded: bool,
+        user: User | None = None,
+        headers: CIMultiDict | None = None,
```
redlickigrzegorz, line 27: It looks like more attributes can be moved to the base `Request` class as they are the same in both classes 🤔 

I mean `context`, `user`, `headers`.

lbz/websocket/request.py
```diff
@@ -0,0 +1,56 @@
...
+class WebSocketRequest(_Request):
+    """Represents request from Web Socket Secure API Gateway.
+
+    rel: https://docs.aws.amazon.com/apigateway/latest/developerguide/
+    apigateway-websocket-api-mapping-template-reference.html
+    """
+
+    def __init__(
+        self,
+        body: str | bytes | dict,
```
redlickigrzegorz, line 22: Maybe it would be good to force using keyword arguments here?
```suggestion
        *,
        body: str | bytes | dict,
```

lbz/rest/request.py
```diff
@@ -1,21 +1,18 @@
...
 logger = get_logger(__name__)
 
 
-class Request:
-    """Represents request from API gateway."""
+class HTTPRequest(_Request):
+    """Represents request from HTTP API Gateway."""
 
     def __init__(
         self,
```
redlickigrzegorz, line 18: The same here:
```suggestion
        self,
        *,
```

lbz/websocket/request.py
```diff
@@ -0,0 +1,56 @@
...
+from lbz.websocket.enums import ActionType
+
+logger = get_logger(__name__)
+
+
+class WebSocketRequest(_Request):
+    """Represents request from Web Socket Secure API Gateway.
+
+    rel: https://docs.aws.amazon.com/apigateway/latest/developerguide/
+    apigateway-websocket-api-mapping-template-reference.html
```
redlickigrzegorz, line 17: You should not split the URL in the middle:
```suggestion
    https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-mapping-template-reference.html
```

lbz/websocket/request.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __init__(
+        self,
+        body: str | bytes | dict,
+        request_details: dict,
+        context: dict,
+        is_base64_encoded: bool,
+        user: User | None = None,
+        headers: CIMultiDict | None = None,
+    ):
```
redlickigrzegorz, line 28:
```suggestion
    ) -> None:
```

lbz/websocket/request.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
+
+from multidict import CIMultiDict
+
+from lbz._request import _Request
+from lbz.authentication import User
+from lbz.misc import get_logger
+from lbz.websocket.enums import ActionType
+
+logger = get_logger(__name__)
```
redlickigrzegorz, line 10: It is not used here as well:
```suggestion
from lbz.websocket.enums import ActionType
```

tests/test_websocket/test_request.py
```diff
@@ -0,0 +1,97 @@
...
+    def test__decode_base64_bytes(self) -> None:
+        encoded = b"asdasdsd"
+        output = WebSocketRequest._decode_base64(encoded)  # pylint: disable=protected-access
+        assert output == b"j\xc7Z\xb1\xdb\x1d"
+
+    def test__decode_base64_str(self) -> None:
+        encoded = "asdasdsd"
+        output = WebSocketRequest._decode_base64(encoded)  # pylint: disable=protected-access
+        assert output == b"j\xc7Z\xb1\xdb\x1d"
+
```
redlickigrzegorz, line 16: You should not test protected methods. This is used by the `raw_body` property and is tested there:
```suggestion
```

tests/test_websocket/test_request.py
```diff
@@ -0,0 +1,97 @@
...
+    @pytest.mark.parametrize(
+        "action_type, is_connection_req",
```
redlickigrzegorz, line 18: Let's name it according to the usage:
```suggestion
        "action_type, expected_result",
```

tests/test_websocket/test_request.py
```diff
@@ -0,0 +1,97 @@
...
+                "stage": "prod",
+            },
+            context={},
+            is_base64_encoded=False,
+        )
+
+        assert request.is_connection_request() == is_connection_req
+
+    @pytest.mark.parametrize(
+        "action_type, is_connection_req",
```
redlickigrzegorz, line 44: ditto:
```suggestion
        "action_type, expected_result",
```

tests/test_websocket/test_request.py
```diff
@@ -0,0 +1,97 @@
...
+                "connectionId": "aaa123",
+                "domainName": "xxx.com",
+                "stage": "prod",
+            },
+            context={},
+            is_base64_encoded=False,
+        )
+        assert request.json_body == {"x": "t1"}
+
+    def test__json_body___returns_dict_when_body_is_a_str(self) -> None:
```
redlickigrzegorz, line 84: Based on the implementation, you should also confirm that:
- `json_body` returns `None` when the body is set to `None`
- `json_body` raises an error when the body is not a JSON-compatible string

redlickigrzegorz (approved):

version
```diff
@@ -1 +1 @@
-0.6.6
+0.7.0
```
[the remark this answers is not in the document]
redlickigrzegorz: It also gives us the possibility to merge everything into master and don't worry about anything 😉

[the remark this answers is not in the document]
pdyba: I wanted to avoid that there is no benefit of releasing unfished code and also we don't want to create a 2K+ LoC PR...

lbz/websocket/request.py
pdyba: headers are optional in WSS (only used on connect) User same case, context is a different case
it might become more clear later but in general Message Event is quite different from Connect/Discconet ones.

pdyba: After some testing I moved the User and context but for headers its beneficial to have them on the current level as they will be always present in REST and Optional in WSS and having them always optional messes with typing.

lbz/websocket/request.py
pdyba: Done - manually - had to check few things.

version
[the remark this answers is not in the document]
pdyba: I would keep for now the current way with a new version branch - ATM there is not that much braking code (import change). And it if in rc3 wie will make a change that will not be compatible with rc1 it doubles our work if rc1 was used anywhere.

tests/test_websocket/test_request.py
pdyba: I've missed that comment I will add that one soon.

lbz/rest/request.py
```diff
@@ -29,45 +27,22 @@ def __init__(
...
+            body=body,
+            is_base64_encoded=is_base64_encoded,
+            context=context,
+            user=user,
+        )
         self.headers = headers
+        self.query_params = MultiDict(query_params or {})
         self.uri_params = uri_params
         self.method = method
         self.context = context
```
redlickigrzegorz, line 40: This line should be removed after the changes you made.

tests/test_websocket/test_request.py
```diff
@@ -0,0 +1,119 @@
...
+                "stage": "prod",
+            },
+            context={},
+            is_base64_encoded=False,
+        )
+        assert request.json_body == {"x": "t1"}
+
+    def test__json_body___handles_none_as_none(self) -> None:
+        request = WebSocketRequest(
+            body=None,  # type: ignore
```
redlickigrzegorz, line 92: It looks like we should allow passing `body=None` instead of ignoring the error.

tests/test_websocket/test_request.py
```diff
@@ -0,0 +1,119 @@
...
+                "connectionId": "aaa123",
+                "domainName": "xxx.com",
+                "stage": "prod",
+            },
+            context={},
+            is_base64_encoded=False,
+        )
+        assert request.json_body is None
+
+    def test__json_body___malformated_json(self) -> None:
```
redlickigrzegorz, line 105:
```python
def test__json_body__raises_error_when_json_is_malformed(self) -> None:
```

lbz/rest/request.py
redlickigrzegorz: FYI: #105

tests/test_websocket/test_request.py
redlickigrzegorz: FYI: #105

tests/test_websocket/test_request.py
redlickigrzegorz: FYI: #105

PR #108, opened 2025-04-11
refactor: using set in return of all_lbz_errors call

allrob23:
This PR updates the `test__all_lbz_errors__returns_all_subclasses_of_lambda_fw_exception` test to use a `set` instead of a `list` for the `result of all_lbz_errors()`. The change optimizes membership checks (using the in operator) by improving lookup performance from O(n) to O(1). Its not a big optimization, but it makes the checks a bit more efficient

The behavior of the test remains unchanged.

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -73,7 +73,7 @@ def test__custom_exception__contains_message_and_status_code_according_to_its_do
 
 
 def test__all_lbz_errors__returns_all_subclasses_of_lambda_fw_exception() -> None:
-    result = list(all_lbz_errors())
+    result = set(all_lbz_errors())
 
     assert BadRequestError in result
     assert ServerError in result


redlickigrzegorz (approved):
It almost does not matter for such a small test suite 🙂 but you are completely right that we should pay more attention to such details in the future, just to simplify our life - thus I approve that 👍

redlickigrzegorz:
But I am not 100% sure why this PR did not trigger our pipelines 🤔 

I will put a pin on that and return when I have more time. Unless you @allrob23 handle that 😉

allrob23:
Thanks for approving. I'm sorry but I don't know how to fix this problem of not running the pipeline.

redlickigrzegorz:
> Thanks for approving. I'm sorry but I don't know how to fix this problem of not running the pipeline.

No problem, I will handle that one day 😉

PR #126, opened 2026-04-03
Do not treat the client-side errors as real warnings worth noting

redlickigrzegorz:
This isn't a huge PR, but it's very useful because with the changes below, we'll stop flooding monitoring systems with warnings that aren't relevant from a backend/server perspective 👏 It's worth noting that I didn't remove all the warnings generated by Lambdalizer, nor the client-sde errors. Almost everything is accessible as it was till now, but on demand, i.e., not always and not by default. In simple words, I let the developers decide what and when they want 😉 

If we need to make it more configurable, we can, but I don't see such a requirement at the moment.


pdyba (approved):

lbz/resource.py
```diff
@@ -63,18 +64,17 @@ def __call__(self) -> Response:
...
             self.response = endpoint(**self.path_params)
-        except LambdaFWException as err:
-            if 500 <= err.status_code < 600:
-                logger.exception(err)
-            else:
-                logger.warning(err, exc_info=is_in_debug_mode())
+        except LambdaFWClientException as err:
+            logger.debug(err, exc_info=True)
+            self.response = Response.from_exception(err, self.request.context["requestId"])
+        except LambdaFWServerException as err:
```
pdyba, line 76: reading that on small screen was problematic ;P

